PR #169, opened 2023-11-12
add version to app.py routes

miararoy:
## Problem

Versioning an API is a crucial aspect of maintaining its usability and compatibility as it evolves over time. There are several approaches to API versioning but the most common is URI Versioning (Path Versioning): where the version number is included in the URL path. For example: 

```
GET /v1/items
POST /v2/items
```

## Solution

- added API_VERSION const to app.py (we could think about a better place to manage this in the future
- changed all routes
- removed /context from chat/completions

> Note:  did not change /health and /shutdown

## Type of Change

## Test Plan

Regression


src/canopy_server/app.py
igiloh-pinecone, line 1: @miararoy You need to change the default `url` paths of all CLI commands that talk to the server (`canopy chat`, `canopy health` etc).  
Please do so after #170 is merged, otherwise you'll have conflicts.

src/canopy_server/app.py
```diff
@@ -83,7 +85,7 @@
 
 
 @app.post(
-    "/context/chat/completions",
```
igiloh-pinecone, line 86: Again, you need to change that in `cli.py` as well (but wait for #170 to merge first!)

src/canopy_server/app.py
```diff
@@ -83,7 +85,7 @@
...
+    f"/api/{API_VERSION}/chat/completions",
```
igiloh-pinecone, line 88: Personally I find this `/api` redundant.  
Up to you, but I vote against it.

tests/e2e/test_app.py
```diff
@@ -95,7 +95,9 @@ def test_health(client):
 
 def test_upsert(client):
     # Upsert a document to the index
-    upsert_response = client.post("/context/upsert", json=upsert_payload.dict())
+    upsert_response = client.post(
+        f"/api/{API_VERSION}/context/upsert",
```
igiloh-pinecone, line 99: That kinda misses the point... 

You want to pass a `base_url` arg to `TestClient()`, and leave the test cases themselves untouched.

src/canopy_server/app.py
```diff
@@ -184,7 +186,7 @@ async def upsert(
 
 
 @app.post(
-    "/context/delete",
+    f"/api/{API_VERSION}/context/delete",
     response_model=SuccessDeleteResponse,
     responses={500: {"description": "Failed to delete documents"}},
 )
```
igiloh-pinecone, line 192: @miararoy if you're adding versioning - than **all** API endpoints should have version, including `/health` etc (it has a structured return value, that may change between versions)

igiloh-pinecone (changes requested):
@miararoy see a few suggestions

src/canopy_server/app.py
miararoy: done

tests/e2e/test_app.py
miararoy: fixed

src/canopy_server/app.py
miararoy: removed

src/canopy_server/app.py
miararoy: Think im' done, resolving this for the future review

src/canopy_cli/cli.py
```diff
@@ -580,8 +582,9 @@ def start(host: str, port: str, reload: bool,
         """
     )
 )
-@click.option("url", "--url", default="http://localhost:8000",
-              help="URL of the Canopy server to use. Defaults to http://localhost:8000")
+@click.option("url", "--url", default=f"http://localhost:8000/{API_VERSION}",
```
igiloh-pinecone, line 585: Let's define a `DEFAULT_URL` at the top of the file?  
It suddenly makes much more sense

miararoy: done

src/canopy_cli/cli.py
```diff
@@ -526,8 +529,8 @@ def chat(chat_server_url, rag, debug, stream):
         """
     )
 )
-@click.option("--host", default="0.0.0.0",
-              help="Hostname or ip address to bind the server to. Defaults to 0.0.0.0")
+@click.option("--host", default="localhost",
+              help="Hostname or address to bind the server to. Defaults to localhost")
```
igiloh-pinecone, line 533: This particular change wasn't tested on all OSes (passing `host='localhost'` to `uvicorn.run()`).  
If you wanna merge it, we will need to test it.

miararoy: does 0.0.0.0 works?

miararoy: on all OSs - that is

src/canopy_server/models/v1/api_models.py
```diff
@@ -4,6 +4,8 @@
 
 from canopy.models.data_models import Messages, Query, Document
 
+# TODO: consider separating these into modules: Chat, Context, Application, etc.
```
igiloh-pinecone, line 7: 👍

igiloh-pinecone (approved):
@miararoy LGTM, but see [this comment](https://github.com/pinecone-io/canopy/pull/169/files#r1392634940) before merging

src/canopy_cli/cli.py
igiloh-pinecone: If that was the code in `main` before this PR - then yeah. I tested this code on all 3 OSes manually yesterday.

PR #170, opened 2023-11-13
CLI and server bug fixes on Windows

### LINKED ISSUE #166: CLI Chat wont start
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

Start canopy, switch to new terminal to run "canopy chat" and it says that canopy isnt running

### Expected Behavior

Canopy chat starts

### Steps To Reproduce

Win environment
Occurs inside and outside virtual environment

### Relevant log output

```shell
Starting Canopy server on 0.0.0.0:8000
INFO:     Started server process [6148]
INFO:     Waiting for application startup.
2023-11-09 20:47:46,231 - MainProcess - canopy_server.app [INFO    ]:  Did not find config file. Initializing engines with default configuration
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)

At same time:
Error: Canopy server is not running on http://0.0.0.0:8000.
please run `canopy start`
```

### Environment

```markdown
- Win:
- python3
```

### Additional Context

I am not a coder, just hobbyist, so likely I am just missing something

igiloh-pinecone:
## Problem

Fixes multiple bugs when running the CLI on windows:
1. On windows, calling CLI commands that use the server (e.g. `canopy chat`) didn't work. 
2. There was also a bug in inferring the `/chat` endpoint, which was using `os.path.join()`
3. `canopy stop` tried to `pgrep gunicorn`, even though Windows doesn't have a `pgrep` command and Gunicorn isn't supported on windows.
4. The `/shutdown` endpoint wasn't working correctly on Windows

Resolves #166. 

## Solution

- The CLI used the url `http://0.0.0.0:8000`/ by default, which is not supported on windows. I replaced it with `http://localhost:8000`.
- In windows we need to send the `CTRL_C` signal, which has a totally different implementation than `SIGINT`

## Type of Change

## Test Plan

Tested manually on mac, Windows and Linux.  
We need to add a proper CLI unit test (known issue)

### DIFF
--- a/src/canopy_cli/cli.py
+++ b/src/canopy_cli/cli.py
@@ -488,7 +488,7 @@ def chat(chat_server_url, rag, debug, stream):
             history=history_with_pinecone,
             message=message,
             stream=stream,
-            api_base=os.path.join(urljoin(url, "/context")),
+            api_base=os.path.join(urljoin(chat_server_url, "/context")),
             print_debug_info=debug,
         )
 
@@ -549,7 +549,7 @@ def start(host: str, port: str, reload: bool,
         "gunicorn canopy_server.app:app --worker-class uvicorn.workers.UvicornWorker "
         f"--bind {host}:{port} --workers <num_workers>"
     ) if os.name != "nt" else (
-        "please use Docker"  #  TODO: Add Docker instructions once we have a Dockerfile
+        "please use Docker"  # TODO: Add Docker instructions once we have a Dockerfile
     )
     for c in note_msg + msg_suffix:
         click.echo(click.style(c, fg="red"), nl=False)
--- a/src/canopy_cli/cli.py
+++ b/src/canopy_cli/cli.py
@@ -171,8 +171,8 @@ def cli(ctx):
 
 
 @cli.command(help="Check if canopy server is running and healthy.")
-@click.option("--url", default="http://0.0.0.0:8000",
-              help="Canopy's server url. Defaults to http://0.0.0.0:8000")
+@click.option("--url", default="http://localhost:8000",
+              help="Canopy's server url. Defaults to http://localhost:8000")
 def health(url):
     check_server_health(url)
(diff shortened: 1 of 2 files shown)


src/canopy_cli/cli.py
```diff
@@ -488,7 +488,7 @@ def chat(chat_server_url, rag, debug, stream):
             history=history_with_pinecone,
             message=message,
             stream=stream,
-            api_base=os.path.join(chat_server_url, "context"),
+            api_base=os.path.join(urljoin(chat_server_url, "/context")),
```
miararoy, line 491: why we still need `os.path.join`

src/canopy_cli/cli.py
```diff
@@ -541,12 +541,17 @@ def start(host: str, port: str, reload: bool,
...
-        "For debugging only. To run the Canopy server in production, run the command:"
+        "For debugging only. To run the Canopy server in production "
+    )
+    msg_suffix = (
+        "run the command:"
         "\n"
         "gunicorn canopy_server.app:app --worker-class uvicorn.workers.UvicornWorker "
         f"--bind {host}:{port} --workers <num_workers>"
+    ) if os.name != "nt" else (
+        "please use Docker"  # TODO: Add Docker instructions once we have a Dockerfile
```
miararoy, line 552: extend here more, "for windows users, inline server experience is not stable, we recommend using docker"

miararoy (changes requested):
small changes

src/canopy_cli/cli.py
igiloh-pinecone: We don't, that's a bug 🤦 .
FIxed

miararoy (approved):
lgtm

PR #171, opened 2023-11-14
Upgrade openai sdk to v.1.2.3

acatav:
## Problem

Canopy is not compatible to new major version change in OpenAI sdk

## Solution

Make Canopy compatible with the new sdk version. Canopy is very new so this PR remove the backwards compatibility 

## Type of Change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Test Plan

Adding tests before making it a PR


src/canopy/llm/openai.py
```diff
@@ -164,17 +159,24 @@ def enforced_function_call(self,
 
         messages = [m.dict() for m in messages]
 
-        chat_completion = openai.ChatCompletion.create(
+        # this enforces the model to call the function
+        function_call = completion_create_params.ChatCompletionFunctionCallOptionParam(
+            name=function.name)
```
igiloh-pinecone, line 164: From my understanding of the [documentation](https://github.com/openai/openai-python/tree/main#nested-params), you're not supposed to instantiate these objects and pass them to the API. I don't see them instantiating typed objects anywhere in their examples.  
You pass a regular `dict`, and they enforce the typing behind the scene (raising a runtime error if the type is wrong).

src/canopy/llm/openai.py
```diff
@@ -164,17 +159,24 @@ def enforced_function_call(self,
...
+
+        chat_completion = self._client.chat.completions.create(
             model=self.model_name,
             messages=messages,
-            functions=[function.dict()],
+            functions=[completion_create_params.Function(
+                name=function.name,
+                description=function.description,
+                parameters=function.parameters.dict())],
             function_call=function_call,
```
igiloh-pinecone, line 173: This field also seems to be deprecated in favor of `tool_choice`, isn't it?

src/canopy/llm/openai.py
```diff
@@ -164,17 +159,24 @@ def enforced_function_call(self,
...
-        chat_completion = openai.ChatCompletion.create(
+        # this enforces the model to call the function
+        function_call = completion_create_params.ChatCompletionFunctionCallOptionParam(
+            name=function.name)
+
+        chat_completion = self._client.chat.completions.create(
             model=self.model_name,
             messages=messages,
-            functions=[function.dict()],
+            functions=[completion_create_params.Function(
```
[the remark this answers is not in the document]
igiloh-pinecone: Also - as above, I think you're supposed to just pass a dict

src/canopy/llm/openai.py
```diff
@@ -30,20 +30,18 @@ def __init__(self,
                  model_name: str = "gpt-3.5-turbo",
                  *,
                  model_params: Optional[ModelParams] = None,
```
igiloh-pinecone, line 32: We probably need to update our `ModelParams` class as well. Please take a look and let's come up with a design. Optional solutions:
- Hard-coded set of params, which we believe that are common to all LLMs. Don't allow other params even if OpenAI has them.
- Hard-coded set of params, but we allow "kwargs" (pydantic extra or similar)
- Free dictionary. Might be problematic if we need to "translate" between external API and underlying models.

src/canopy/llm/openai.py
```diff
@@ -164,17 +159,24 @@ def enforced_function_call(self,
...
             messages=messages,
-            functions=[function.dict()],
+            functions=[completion_create_params.Function(
+                name=function.name,
+                description=function.description,
+                parameters=function.parameters.dict())],
             function_call=function_call,
             max_tokens=max_tokens,
             **model_params_dict
         )
```
igiloh-pinecone, line 176: From the docs - I think you also need to pass `response_format={ "type": "json_object" }`.  
Or is that guaranteed \ automatic if `function_call` ( or `tool_choice`) is True?

src/canopy/models/data_models.py
```diff
@@ -112,3 +112,11 @@ class SystemMessage(MessageBase):
 class AssistantMessage(MessageBase):
     role: Literal[Role.ASSISTANT] = Role.ASSISTANT
     content: str
+
+
+class OpenAIClientParams(BaseModel):
```
igiloh-pinecone, line 117: This class definitely doesn't belong here.  
It's an inner config-level detail of a specific LLM (theoretically out of many supported LLMs). The `data_models` file is for our main API data classes like `Query` and `Context` etc.  

I think you can safely move it inside the `llm/openai.py` file.

src/canopy/models/data_models.py
```diff
@@ -112,3 +112,11 @@ class SystemMessage(MessageBase):
...
+    api_key: Optional[str] = None
+    organization: Optional[str] = None
+    base_url: Optional[str] = None
+    max_retries: Optional[int] = 3
+    timeout: Optional[int] = 30
```
igiloh-pinecone, line 122: Are we absolutely sure that we want retry to be part of client params?  
Isn't it an `LLM` param of its own?

(Maybe it's not, just asking)

src/canopy/llm/openai.py
```diff
@@ -164,17 +159,24 @@ def enforced_function_call(self,
...
 
-        result = chat_completion.choices[0].message.function_call
-        arguments = json.loads(result["arguments"])
+        result = chat_completion.choices[0].message.function_call  # type: ignore
```
igiloh-pinecone, line 178: Why did you need to `type: ignore` here?  
This seems suspicious. Are you sure we're not missing an actual required conversion (and\or something with the `response_format` arg)?

igiloh-pinecone (changes requested):
@acatav Please read the comments, then maybe let's do a quick sync to discuss?

src/canopy/llm/openai.py
```diff
@@ -3,17 +3,17 @@
 import jsonschema
 import openai
 import json
+
+from openai.types.chat import completion_create_params
```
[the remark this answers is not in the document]
igiloh-pinecone: Please see below about the `TypedDict` thing - I'm not convinced we even need this function at all

src/canopy/knowledge_base/record_encoder/openai.py
```diff
@@ -34,15 +29,10 @@ def __init__(self,
                         Defaults to 400.
             **kwargs: Additional arguments to pass to the underlying `pinecone-text. OpenAIEncoder`.
         """  # noqa: E501
-        encoder = OpenAIEncoder(model_name)
+        client_params = client_params or OpenAIClientParams()
```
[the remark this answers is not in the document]
acatav: changed

src/canopy/llm/openai.py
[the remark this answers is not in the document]
acatav: changed

src/canopy/llm/openai.py
acatav: changed

src/canopy/llm/openai.py
acatav: changed

src/canopy/llm/openai.py
acatav: we chose free dict

src/canopy/models/data_models.py
acatav: removed

src/canopy/llm/openai.py
acatav: we used cast

src/canopy/models/data_models.py
acatav: removed

pyproject.toml
```diff
@@ -30,6 +29,7 @@ types-pyyaml = "^6.0.12.12"
 jsonschema = "^4.2.0"
 types-jsonschema = "^4.2.0"
 prompt-toolkit = "^3.0.39"
+pinecone-text = {git = "https://github.com/pinecone-io/pinecone-text.git"}
```
acatav, line 32: change to new version

pyproject.toml
igiloh-pinecone, line 1: Missing changes:

1. Change the notebook
2. Change the library readme
3. Verify CLI

igiloh-pinecone (commented):
A few comments

igiloh-pinecone (approved):
LGTM

PR #178, opened 2023-11-16
A few bug fixes following OpenAI client update

igiloh-pinecone:
## Problem

After OpenAI client was updated (#171), a few problems arose:
- Running the `canopy` CLI command (without subcommand) always errored out
- `OpenAILLM.avilable_models` was broken

## Solution
There was a minor bug in the CLI that still used the old `openai` syntax.  
But there was a deeper issue: Since we'll be adding support for more LLMs soon, it doesn't make sense that the CLI verifies the OpenAI connection at the top level. Changes:
- Each CLI sub-command now verifies it own dependencies (e.g. connection with Pinecone)
- Connection to OpenAI is only checked if an `OpenAI` component is in the configuration.
- I added a dedicated config verification step in `canopy start`, before running the server itself. This allows the CLI to catch the actual error, and present it to the user in a controlled manner (instead of the error being raise inside the `UVicorn` process itself, which we can't catch).
- Fixed the bug in `OpenAILLM.availble_models` and add a UT

## Type of Change

## Test Plan

- Add missing test case to `OpenAILLM`
- We're still missing a proper CLI test, which we should be adding ASAP

### DIFF
--- a/src/canopy_cli/cli.py
+++ b/src/canopy_cli/cli.py
@@ -84,7 +84,8 @@ def validate_connection():
         )
         raise CLIError(msg)
     try:
-        openai.Model.list()
+        client = openai.OpenAI()
+        client.models.list()
     except Exception:
         msg = (
             "Failed to connect to OpenAI, please make sure that the OPENAI_API_KEY "
--- a/src/canopy_cli/cli.py
+++ b/src/canopy_cli/cli.py
@@ -144,7 +144,7 @@ def _validate_chat_engine(config_file: Optional[str]):
     except openai.OpenAIError:
         raise CLIError(OPENAI_AUTH_ERROR_MSG)
     except Exception as e:
-        msg = f"Failed to initialize chat engine. Reason:\n{e}"
+        msg = f"Failed to initialize Canopy server. Reason:\n{e}"
         if config_file:
             msg += f"\nPlease check the configuration file {config_file}"
         raise CLIError(msg)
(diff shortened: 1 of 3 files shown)


src/canopy_cli/cli.py
```diff
@@ -104,17 +100,25 @@ def _initialize_tokenizer():
...
     except Exception as e:
         msg = f"Failed to load config file {config_file}. Reason:\n{e}"
         raise CLIError(msg)
 
+    return config
+
+
+def _load_kb_config(config_file: Optional[str]) -> Dict[str, Any]:
+    config = _read_config_file(config_file)
+    if not config:
```
miararoy, line 119: Isn't it a duplicate?

src/canopy_cli/cli.py
```diff
@@ -132,6 +136,22 @@ def _load_kb_config(config_file: Optional[str]) -> Dict[str, Any]:
     return kb_config
 
 
+def _validate_chat_engine(config_file: Optional[str]):
+    config = _read_config_file(config_file)
+    Tokenizer.initialize()
```
miararoy, line 141: We should have a "with tokenizer..."
A nice feature

igiloh-pinecone: Yeah, thought about it.  
Doesn't worth the effort right now...

src/canopy_cli/cli.py
igiloh-pinecone: Nope, `config` can return empty (`{}`) from `_read_config_file()`.

src/canopy_cli/cli.py
miararoy: Just task, i will add

miararoy (approved):
Lg

PR #180, opened 2023-11-16
Added release 0.2.0 to CHANGELOG

igiloh-pinecone:
The version bump itself would happen in the release workflow, which is finally fixed!

(In the near future we will switch to automatic CHANGELOG)


miararoy (approved):
Lg

Issue #181, opened 2023-11-17
[Bug] Uninformative error message when upserting files where UTF-8 file decoding fails


mjul:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

When importing a file with canopy-cli (upsert) the `f.read()` in `canopy_cli\data_loader\data_loader.py` can throw a `UnicodeDecodeError` exception. 

This results in the CLI responding with a very uninformative error message like this:

    Error: A unexpected error while loading the data from files in .\data\hello\test.txt. Please make sure the data is in valid `jsonl`, `parquet`, `csv` format or plaintext `.txt` files.

If I patch the code to print out the exception like this, 

            try:
                text = f.read()
            except UnicodeDecodeError as e:
                print("Error reading file: ", e)

It prints the exception message: 

    'charmap' codec can't decode byte 0x9d in position 249: character maps to <undefined>

So, it decoding UTF-8 seems to be the root problem.

### Expected Behavior

Either import the file, handling the error, or provide a better error message so the files can be fixed.

### Steps To Reproduce

I am using Python 3.11 on Windows.

To reproduce create a text file `test.txt` as a UTF-8 encoded text file containing e.g. a 0x9d byte.

`canopy upsert test.txt` then fails:

    Error: A unexpected error while loading the data from files in test.txt. Please make sure the data is in valid `jsonl`, `parquet`, `csv`    format or plaintext `.txt` files.

### Environment

```markdown
- **OS: Windows 11**:
- **Language version: Python 3.11.5**:
- **Canopy version: canopy-sdk==0.1.4**:
```

mjul:
If i change the file `open` in `data_loader.py` line 93 to specify the encoding, e.g. `        with open(file_path, "r", encoding="utf-8") as f:` I can make it read the UTF-8 encoded files, so it appears to somehow select another encoding than UTF-8 for the txt file it is reading.

miararoy:
Hi @mjul !
Thanks for flagging, we internally know our data loading error handling is lacking and I am actively working on a PR for this.

I think utf-8 is good enough for now.

as the PR handles more cases - our specific solution here would be something like:

```python
for file_path in file_paths:
        try:
            with open(file_path, "r", encoding='utf-8') as f:
                text = f.read()
                rows.append(
                    {
                        "id": os.path.basename(file_path).replace(".txt", ""),
                        "text": text,
[...]

PR #182, opened 2023-11-19
Improve dataloader error handling

miararoy:
## Problem

data loading has no meaningful error handling, albeit being a sensitive part of the flow

resolves: #181

## Solution

first, defining `DataLoaderException` that will formalize all data loading error on both the file/row level and unify them under one error, then change the exception handling to raise this error

## Type of Change

## Test Plan

- adding simple tests for data loading of .txt 
- changing the current mechanism of dataloader tests to also have the data as (name, df, exception) for the bad cases

note that tests should were added to txt -> df as from df to documents is handled.


src/canopy_cli/data_loader/errors.py
```diff
@@ -0,0 +1,37 @@
...
+
+    #: The exit code for this exception.
+    exit_code = 1
+
+    def __init__(self, file_name: str, row_id: str, err: str) -> None:
+        message = f"""
+        [ERROR] {file_name} - {row_id} - {err}
+        """
+        super().__init__(message)
+        self.message = message
```
igiloh-pinecone, line 25: That kinda breaks the whole concept of a structured error. 

Save each one of the fields as its own member variable (`self.file_name` etc), then only format the string at `format_message()`. This one someone can `except DataLoaderException as e`  and use the structured information somehow.

src/canopy_cli/data_loader/errors.py
```diff
@@ -0,0 +1,37 @@
...
+
+class DataLoaderException(Exception):
+    """An exception that Click can handle and show to the user."""
+
+    #: The exit code for this exception.
+    exit_code = 1
+
+    def __init__(self, file_name: str, row_id: str, err: str) -> None:
+        message = f"""
+        [ERROR] {file_name} - {row_id} - {err}
```
igiloh-pinecone, line 22:
```suggestion
        {file_name}, line {row_id} - {err}
```

No need for the `Error` - the CLI adds this prefix anyway

src/canopy_cli/data_loader/errors.py
```diff
@@ -0,0 +1,37 @@
...
+        return self.message
+
+    def __str__(self) -> str:
+        return self.message
+
+    def show(self, file: t.Optional[t.IO] = None) -> None:
+        if file is None:
+            file = get_text_stderr()
+
+        echo("Error: {message}".format(message=self.format_message()), file=file)
```
igiloh-pinecone, line 37:
```suggestion
        echo("{message}".format(message=self.format_message()), file=file)
```
No need for the `Error:`, the CLI adds that anyway

src/canopy_cli/data_loader/data_loader.py
```diff
@@ -61,10 +57,20 @@ def _df_to_documents(df: pd.DataFrame) -> List[Document]:
...
+        documents = []
+        for row in df.itertuples(index=False):
+            try:
+                documents.append(
+                    Document(
+                        **{k: v for k, v in row._asdict().items() if not pd.isna(v)}
+                    )
+                )
+            except ValidationError as e:
+                raise DataLoaderException(
```
igiloh-pinecone, line 69: You're still `raise`ing on the first error, always.

We should add an `allow_errors` bool and a tolerance level (can be hard-coded, say 5% of the data). If `allow_errors=True`, errors are ignored if less than the allowed tolerance (number of docs out of the total `df`) have failed.

src/canopy_cli/data_loader/data_loader.py
```diff
@@ -61,10 +57,20 @@ def _df_to_documents(df: pd.DataFrame) -> List[Document]:
...
-        ]
+        documents = []
+        for row in df.itertuples(index=False):
+            try:
+                documents.append(
+                    Document(
+                        **{k: v for k, v in row._asdict().items() if not pd.isna(v)}
+                    )
+                )
+            except ValidationError as e:
```
igiloh-pinecone, line 68: If you don't have a dedicated handling of `ValidationError` which is different than how you'd handle other error types - just stick with `except Exception as e`.  
Otherwise we might miss certain Exceptions, and they will halt execution ungracefully.

igiloh-pinecone (changes requested):
@miararoy see a few suggestions please

src/canopy_cli/data_loader/errors.py
miararoy: yeah makes sense!

src/canopy_cli/data_loader/data_loader.py
miararoy: Sounds like a new feature? we can start by failing every time and then build a strategy according to user feedback, do you think we should prioritize it now?

src/canopy_cli/data_loader/data_loader.py
miararoy: I intended other errors that are not validation to be raised as is (and with time handle them) note that the only think we are trying is to construct a document. does it make sense?

src/canopy_cli/data_loader/errors.py
miararoy: done

src/canopy_cli/data_loader/data_loader.py
igiloh-pinecone: If you'll look at `Cli.upsert()`, we're already doing that for upsert errors. I don't see a reason not to do the same for loader errors (it's literally the same, from the user's perspective).

We can defer this to a separate PR, if you prefer (just please make sure we have an issue tracking this effort somewhere)

miararoy: I think we should 

Opened a task

https://app.asana.com/0/1204545509946596/1202897920923848

igiloh-pinecone (approved):
LGTM

PR #187, opened 2023-11-25
add document field explanations and python version badges

### LINKED ISSUE #186: [RAG meta-data] Include meta-data of the RAG retrieved documents
### Is this your first time submitting a feature request?

- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing functionality

### Describe the feature

As far as I can see, for the RAG generated response, there's no meta-data with regards to which vector documents it has used. This would be very useful for end-users so that they can get references as to which documents the response is referring to.

**Edit** Having dived into the code base, is this intended to be build perhaps by clients using this framework by building their own ContextBuilder?

### Describe alternatives you've considered

_No response_

### Who will this benefit?

I believe there's a lot use cases that would want to prove that what they output is correct. The end user can manually verify or get more in-depth information based on meta-data coming back from RAG retrieved documents.

### Are you interested in contributing this feature?

My Python knowledge is very limited, but I'll give it a go if maintainers direct me as to how they see the data object structure change for example.

### Anything else?

Nothing at the moment, I think the feature is pretty straight-forward. I'd suggest the client supports a simple list of strings representing the meta-data properties they would like to retrieve from Pinecone, and include those in the RAG response.

miararoy:
## Problem

The readme lacks two important features:
- python support badges and package version
- more explanation about the uses of the fields in the document

## Solution

- change the docs
- close #186 

## Type of Change

- [x] Non-code change (docs, etc)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,15 @@
 # Canopy
 
+<p align="center">
+<a href="https://pypi.org/project/fastapi" target="_blank">
+    <img src="https://img.shields.io/pypi/pyversions/canopy-sdk" alt="Supported Python versions">
+</a>
+<a href="https://pypi.org/project/fastapi" target="_blank">
+    <img src="https://img.shields.io/pypi/v/canopy-sdk?label=pypi%20package" alt="Package version">
+</a>
+</p>
+
+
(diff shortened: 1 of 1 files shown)


igiloh-pinecone (approved):
LGTM!

PR #188, opened 2023-11-27
Azure OpenAI LLM implementation

MichaelAnckaert:
## Problem

Canopy can't be used with the Azure OpenAI offering.

## Solution

This PR creates a new AzureOpenAILLM class (inherits from OpenAILLM) to be used with the Azure OpenAI offering. 

## Type of Change

## Test Plan

No tests added yet.


src/canopy_server/app.py
```diff
@@ -332,7 +332,15 @@ def _init_engines():
         Tokenizer.initialize()
         kb = KnowledgeBase(index_name=index_name)
         context_engine = ContextEngine(knowledge_base=kb)
-        llm = OpenAILLM()
+        if llm == "OpenAILLM":
+            llm = OpenAILLM()
+        elif llm == "AzureOpenAILLM":
+            llm = AzureOpenAILLM(
+                api_key=os.getenv("OPENAI_API_KEY"),
+                base_url=os.getenv("OPENAI_BASE_URL"),
```
miararoy, line 340: The original OpenAI client is automatically gettitng the env, isn't the AzureOpenAIClient doing the same? if so we prefer handling it this way

miararoy: seems it does:
https://github.com/openai/openai-python/blob/main/src/openai/lib/azure.py#L160

src/canopy/llm/openai.py
```diff
@@ -206,3 +207,135 @@ async def agenerate_queries(self,
...
+            api_key: Your OpenAI API key. Defaults to None (uses the "OPENAI_API_KEY" environment variable).
+            organization: Your OpenAI organization. Defaults to None (uses the "OPENAI_ORG" environment variable if set, otherwise uses the "default" organization).
+            base_url: The base URL to use for the OpenAI API. Defaults to None (uses the default OpenAI API URL).
+            **kwargs: Generation default parameters to use for each request. See https://platform.openai.com/docs/api-reference/chat/create
+                    For example, you can set the temperature, top_p etc
+                    These params can be overridden by passing a `model_params` argument to the `chat_completion` or `enforced_function_call` methods.
+        """  # noqa: E501
+        super().__init__(model_name)
+        self._client = openai.AzureOpenAI(
+            api_key=api_key or os.getenv("OPENAI_API_KEY"),
```
miararoy, line 243: Note that the new openaiclient is automatically getting the AZURE_OPENAI_API_KEY param
https://github.com/openai/openai-python/blob/main/src/openai/lib/azure.py#L160

src/canopy_server/app.py
```diff
@@ -332,7 +332,15 @@ def _init_engines():
...
+            )
+        else:
+            raise ValueError("Config issue: LLM must be either OpenAILLM or AzureOpenAILLM")
```
miararoy, line 343: Config loader should automatically handle values, no need to have it here, in fact this will block other LLM implementations

miararoy (changes requested):
Overall direction seems good, main changes is:
- env loading should be handled by client or by explicit passing parameters
- llm object creation in app.py
- adding proper docstrings and adding readme

MichaelAnckaert:
> Overall direction seems good, main changes is:
> 
>     * env loading should be handled by client or by explicit passing parameters
> 
>     * llm object creation in app.py
> 
>     * adding proper docstrings and adding readme

Thanks for your review @miararoy Happy to know the direction is the correct one. 
I'll address your comments and update the PR.

src/canopy/llm/openai.py
```diff
@@ -206,3 +207,135 @@ async def agenerate_queries(self,
...
+            api_version="2023-10-01-preview",
```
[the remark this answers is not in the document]
MichaelAnckaert: From the Azure documentation this should be a date formatted version.
I'll specified a default version here as the OpenAI SDK requires a version to be set, either via parameter or via environment variable. It seemed a bit excessive to me to require the user to specify yet another environment variable. But willing to change if there's another point of view. 

Ref: https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#rest-api-versioning
Ref: https://github.com/openai/openai-python/blob/17ac6779958b2b74999c634c4ea4c7b74906027a/src/openai/lib/azure.py#L170

miararoy:
Thanks! 
Just letting you know we have internally started to work on this as well, we will make sure to merge the work so this could be shipped asap

src/canopy/llm/openai.py
```diff
@@ -206,3 +207,135 @@ async def agenerate_queries(self,
...
+
+        Args:
+            model_name: The name of the model to use. See https://platform.openai.com/docs/models
+            api_key: Your Azure OpenAI API key. Defaults to None (uses the "OPENAI_API_KEY" environment variable).
+            base_url: The base URL to use for the Azure OpenAI API. Will use the AZURE_OPENAI_ENDPOINT environment variable if set.
+            **kwargs: Generation default parameters to use for each request. See https://platform.openai.com/docs/api-reference/chat/create
+                    For example, you can set the temperature, top_p etc
+                    These params can be overridden by passing a `model_params` argument to the `chat_completion` or `enforced_function_call` methods.
+        """  # noqa: E501
+        super().__init__(model_name)
```
igiloh-pinecone, line 240:
```suggestion
        super().__init__(model_name=None)
```
There's not point sending the `model_name` to `super().__init__` and save it under `self.model_name`.  
In the `chat.completion.create()` call, the `model=` argument is interpreted as the `deployment` anyway (and ignored if `deployment` was explicitly defined).  
In the context of `AzureOpenAILLM`, the init arg `model_name` should be synonym with `model deployment`. Of course the docstring would need to explain that carefully.

src/canopy/llm/openai.py
```diff
@@ -206,3 +207,135 @@ async def agenerate_queries(self,
...
+        self._client = openai.AzureOpenAI(
+            api_key=api_key,
+            api_version="2023-10-01-preview",
+            azure_endpoint=base_url,
```
igiloh-pinecone, line 244:
```suggestion
            azure_endpoint=base_url,
            azure_deployment=model_name
```
See comment above about using `self.model_name`

src/canopy/llm/openai.py
```diff
@@ -206,3 +207,135 @@ async def agenerate_queries(self,
...
+
+        model_params_dict: Dict[str, Any] = deepcopy(self.default_model_params)
+        model_params_dict.update(model_params or {})
+
+        function_dict = cast(ChatCompletionToolParam, function.dict())
+
+        chat_completion = self._client.chat.completions.create(
+            model=self.model_name,
+            messages=[m.dict() for m in messages],
+            functions=[function_dict],
```
igiloh-pinecone, line 312: A. Are we sure that if you deploy your model with the latest GPT version (e.g. `gpt-3.5-turbo-1106`), the `tools=` argument is still not supported?  
If it does - we should probably remove the duplication and use `OpenAILLM`'s implementation directly.

igiloh-pinecone: B. I noticed that the default `gpt-3.5-turbo` version on Azure OpenAI services is `0301`, which doesn't support function calling. It means that for many users - this code is going to error out, and OpenAI's error message is not very intuitive.  

We should probably add a `try: except openai.NotFoundError:` clause around this call, and re-raise with a clearer error message that tells the user that the model version defined in their deployment doesn't support function calling.

src/canopy_server/app.py
```diff
@@ -18,12 +18,7 @@
 from starlette.concurrency import run_in_threadpool
 from sse_starlette.sse import EventSourceResponse
 
-from fastapi import (
-    FastAPI,
-    HTTPException,
-    Body,
-    APIRouter
-)
+from fastapi import FastAPI, HTTPException, Body, APIRouter
```
igiloh-pinecone, line 21: @MichaelAnckaert for this PR you can leave these changes as-is, but in the future please avoid running linters like `black` on files (or even lines) that weren't  changed in your PRs. This results in redundant conflicts with other PRs.

src/canopy/llm/openai.py
```diff
@@ -206,3 +207,135 @@ async def agenerate_queries(self,
...
+            max_tokens=max_tokens,
+            **model_params_dict,
+        )
+
+        result = chat_completion.choices[0].message.function_call.arguments
+        arguments = json.loads(result)
+
+        jsonschema.validate(instance=arguments, schema=function.parameters.dict())
+        return arguments
+
```
igiloh-pinecone, line 323: Note that the `.available_models` property is broken, as it seems that Azure OpenAI API doesn't allow listing all of your deployments (==models).  
We should wrap that with a clear error message, and point to where the list of available deployments can be found in the Azure console

src/canopy/knowledge_base/record_encoder/openai.py
```diff
@@ -42,9 +44,57 @@ def encode_documents(self, documents: List[KBDocChunk]) -> List[KBEncodedDocChun
...
-                                       ) -> List[KBEncodedDocChunk]:
+    async def _aencode_documents_batch(
+        self, documents: List[KBDocChunk]
+    ) -> List[KBEncodedDocChunk]:
+        raise NotImplementedError
+
+    async def _aencode_queries_batch(self, queries: List[Query]) -> List[KBQuery]:
+        raise NotImplementedError
+
+
```
igiloh-pinecone, line 55: Last but not least - please note that this class requires its own unit test before it can be merged.

src/canopy/llm/openai.py
```diff
@@ -206,3 +207,135 @@ async def agenerate_queries(self,
...
+        raise NotImplementedError()
+
+    async def agenerate_queries(
+        self,
+        messages: Messages,
+        *,
+        max_generated_tokens: Optional[int] = None,
+        model_params: Optional[dict] = None,
+    ) -> List[Query]:
+        raise NotImplementedError()
```
igiloh-pinecone, line 341: Note that the `AzureOpenAILLM` class requires its own system test (since we don't have stand alone unit test for `LLM` classes yet).  
The test can assume that an Azure OpenAI endpoint was created, and that both the endpoint and the API key are provided by env vars.  
Once the test file exists (and passes locally on your branch) - we will add `AZURE_OPENAI_API_KEY` and `AZURE_OPENAI_ENDPOINT` secretes to the repo's CI configuration.

src/canopy/llm/openai.py
```diff
@@ -206,3 +207,135 @@ async def agenerate_queries(self,
...
+    Azure OpenAI LLM wrapper built on top of the OpenAI Python client.
+
+    Note: OpenAI requires a valid API key to use this class.
+          You can set the "AZURE_OPENAI_KEY" environment variable to your API key.
+          Or you can directly set it as follows:
+          >>> import openai
+          >>> openai.api_key = "YOUR_API_KEY"
+    """
+    def __init__(self,
+                 model_name: str = "gpt-3.5-turbo",
```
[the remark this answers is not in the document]
MichaelAnckaert: Makes me wonder if I should change the implementation of `AzureOpenAILLM` to have different init arguments from the default `OpenAILLM` class? I added the `api_version` argument, does it make sense to rename `model_name` to `deployment_name`?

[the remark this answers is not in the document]
igiloh-pinecone: @MichaelAnckaert `model_name` is part of the signature of all `BaseLLM` inheritors, so that is the only parameter that has to still exist in order to comply with `BaseLLM` interface.  
You can add additional init arguments or even change\remove ones that come from `OpenAILLM`, as long as you maintain compatibility with  `BaseLLM`.

igiloh-pinecone:
@MichaelAnckaert there hasn't been any activity in this PR since the review two weeks ago.  
Please advise if you are planning to finalize this PR so it can be merged.

igiloh-pinecone:
@MichaelAnckaert Thank you very much for your contribution!!

Since this PR has been inactive, we have finalized it with all missing tests and prepared it for merging.  
All of your original commits are included of course, and will be merged to Canopy.

acatav (dismissed):
LGTM overall, look at the minor suggestions. I mainly think we should decide if api version goes to env vars or config, and can consider enable model name in config at some sense

igiloh-pinecone (approved):
LGTM!

Issue #189, opened 2023-11-27
[Bug] Command `canopy new` doesn't read config from file specified by environment variable


MichaelAnckaert:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

When running `canopy new`, the config file specified by the environment variable $CANOPY_CONFIG_FILE is not read. The config file must be manually specified using the `-c` parameter.

### Expected Behavior

The config file specified by $CANOPY_CONFIG_FILE is read.

### Steps To Reproduce

Run `canopy new` 

### Environment

```markdown
- **Canopy version**: master branch
```

igiloh-pinecone:
@MichaelAnckaert thanks for reporting this!

Will be fixed in the next patch release

igiloh-pinecone:
@MichaelAnckaert  sure, go ahead!

igiloh-pinecone:
Resolved by #190

PR #190, opened 2023-11-28
CLI: read config file from env location (Issue #189)

MichaelAnckaert:
This PR fixes the bug described in issue #189.

## Problem

The issue occurs then the user runs a Canopy CLI command (such as `new`, `upsert`) and does not specify the `-c` parameter with a config file lcoation. 
The current behaviour is to default to an empty (default) configuration while the user might expect Canopy to read the config file location specified in the environment variable `CANOPY_CONFIG_FILE`.

## Solution

This change updates the way the config file is loaded by the Canopy CLI. When the config file location is not specified through the `-c` parameter, we use the file location specified in the environment variable CANOPY_COPY_FILE. 

## Type of Change

## Test Plan

Only manual testing was done for this change.

### DIFF
--- a/src/canopy_cli/cli.py
+++ b/src/canopy_cli/cli.py
@@ -102,9 +102,7 @@ def _initialize_tokenizer():
 
 def _read_config_file(config_file: Optional[str]) -> Dict[str, Any]:
     if config_file is None:
-        config_file = os.getenv("CANOPY_CONFIG_FILE", None)
-        if config_file is None:
-            return {}
+        return {}
 
     try:
         with open(config_file, 'r') as f:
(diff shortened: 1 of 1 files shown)


src/canopy_cli/cli.py
```diff
@@ -102,7 +102,9 @@ def _initialize_tokenizer():
 
 def _read_config_file(config_file: Optional[str]) -> Dict[str, Any]:
     if config_file is None:
-        return {}
+        config_file = os.getenv("CANOPY_CONFIG_FILE", None)
```
igiloh-pinecone, line 105: I believe `click` has a built-in solution, that should also automatically appear in the help text.  
If you'll look at the click declaration of the `--index-name` argument of `canopy new`, you'd notice an `envvar=` option that that automatically sets that argument to also be read from an en var.

igiloh-pinecone (approved):
LGTM!

(Have manually fixed PR comments that wasn't fixed for a week)

Issue #164, opened 2023-11-09
[Feature] Allow end users to pass their own LLM API key and LLM model params


igiloh-pinecone:
### Is this your first time submitting a feature request?

- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [ ] I am requesting a straightforward extension of existing functionality

### Describe the feature

We currently designed `canopy-server` and the `LLM` class under the assumption that a developers would build a Chatbot service with Canopy which offers a single bot service. Under that assumption - all the LLM calls are done by a single pre-configured API key, using pre-configured model params (such as `model_name`, `max_generated_tokens`).

However, some developers might want to offer a multi-tenant bot service over the same knowledge base. In that case, the server would need to take the API key and the model params from each individual `/chat.completion` call, and the LLM should allow passing these per call.

### Describe alternatives you've considered

We should probably add a `bool` in the `LLM`'s config which determines whether or not to allow to pass model params per call.  
We can maybe add a separate config flag specifying whether or not to allow users to pass their own personal LLM API keys.

Theoretically, we can even do a more detailed `allow_override` bool per each model param - but that's probably an overkill.

### Who will this benefit?

Developers who want to use Canopy to offer multi-tenant chatbot service.

malexw:
👋 Hi @igiloh-pinecone,

I though I'd take a look at implementing this. Do you have a vision for what this set of changes looks like?

For the most part passing model parameters through looks fairly straightforward to me. Two pain points I see are around handling API keys, or passing model parameters in a world where Canopy supports additional LLMs besides OpenAI. Using the Cohere LLM as an example, their equivalent of `/chat/completions` is just `/generate`. Is the plan for Canopy to expose additional routes in the server that align with the APIs of the other LLMs, or will canopy offer a single combined API route that handles requests that eventually make their way to the appropriate `ChatEngine`?

I notice that both the OpenAI and Cohere python libraries want an API key provided when the client is instantiated. Do you want to go down the path of instantiating appropriate `LLM`s and `ChatEngine`s per-request? Maybe for v1 of these changes it would make sense to continue to use an API key provided by the `config.yaml` or an env var, and just pass the model params?

igiloh-pinecone:
Hi @malexw , thank you for your contribution!!!

The general guidelines we've envisioned so far are:
- Canopy's external API should be OpenAI-like, regardless of the backend LLM class configured (like CohereLLM).
- Only one LLM class should be configured at any given time

[...]

PR #191, opened 2023-11-28
Pass additional model params to the LLM

malexw:
## Problem

Currently `canopy-server` and the `LLM` model params are all pre-configured. In order to offer more flexibility in how the underlying LLM is used, we'd like to enhance `canopy-server` to forward the params that are passed with each `/chat/completion` call to the underlying `LLM`.

## Solution

This PR addresses issue #164. We add two new config parameters, `allow_api_key_override` and `allow_model_params_override`, with both defaulting to `true`.

When `allow_api_key_override` is true, Canopy will read the API key from the request's `Authorization` header, if set, and forward it on to the LLM to be used in API calls. If no `Authorization` header is set or the config value is false, Canopy continues to use the api key specified by the config or env var.

When `allow_model_params_override` is true, any model parameters specified in the API call are forwarded to the LLM. For example, in the OpenAI API, these might be `temperature`, `top_p`, or `logit_bias`. Two parameters that aren't currently forwarded are `tools` and `tool_choice`, as well as the deprecated `functions` and `function_call`. Getting Canopy to work with function calling would require a bit of additional work that I haven't done here.

## Type of Change

- [x] This change requires a documentation update

## Test Plan

Tests were run manually to check that parameters are getting passed through and that Canopy properly handles the response coming back (for example, that all the choices are returned for `n` > 1). Two unit tests were also added one for the `override` config values being `False`, and one for them being `True`.


src/canopy/llm/openai.py
```diff
@@ -94,9 +94,11 @@ def chat_completion(self,
         model_params_dict.update(
             model_params or {}
         )
+        model = model_params.get("model", self.model_name) if model_params else self.model_name
+        model_params_dict = {k: v for k, v in model_params_dict.items() if k != 'model' and v is not None} if model_params else None
```
malexw, line 98: I think I more complete solution here will probably need to make sure we're only passing params to OpenAI that their API supports

src/canopy/chat_engine/chat_engine.py
```diff
@@ -197,8 +197,10 @@ def chat(self,
             messages,
             max_tokens=self.max_prompt_tokens
         )
+        max_tokens = model_params.get("max_tokens", self.max_generated_tokens) if model_params else self.max_generated_tokens
+        model_params = {k: v for k, v in model_params.items() if k != 'max_tokens'} if model_params else None
         llm_response = self.llm.chat_completion(llm_messages,
```
malexw, line 202: Should we be updating `BaseLLM` to move the `max_tokens` and `stream` param inside of `model_params`?

Or should we go the other way and pull more params out of `model_params`? Or neither?

src/canopy/llm/openai.py
```diff
@@ -91,15 +93,19 @@ def chat_completion(self,
...
-        model_params_dict.update(
-            model_params or {}
-        )
+        if self.request_params_passthrough:
+            passthrough_params = {
+                k: v for k, v in (model_params or {}).items() if v is not None
+            }
+            model_params_dict.update(passthrough_params)
+        if model_params_dict.get("model", None) is None:
+            model_params_dict["model"] = self.model_name
```
igiloh-pinecone, line 102: You can probably do that already on `__init()__`:
```python
        super().__init__(model_name)
        ...
        self.default_model_params = kwargs
        self.default_model_params["model] = self.model_name
```
And then the `update()` above would suffice

igiloh-pinecone (changes requested):
@malexw Please see one suggested change - move the passthrough control from the `LLM` level to the `ChatEngine` level.

src/canopy/llm/openai.py
malexw: Good suggestion!

config/config.yaml
```diff
@@ -50,7 +52,7 @@ chat_engine:
     params:
       model_name: gpt-3.5-turbo         # The name of the model to use.
       # You can add any additional parameters which are supported by the LLM's `ChatCompletion()` API. The values set
-      # here will be used in every LLM API call. For example:
+      # here will be used in every LLM API call unless `request_params_passthrough` is false. For example:
```
igiloh-pinecone, line 55:
```suggestion
      # here will be used in every LLM API call unless `allow_model_params_override` is false. For example:
```

malexw:
Sure thing @igiloh-pinecone - I've updated this one to be just for param passing.

The API key changes part of this PR didn't feel right to me either, but without doing deeper rearchitecting I'm not sure that we had a lot of options. Should I hold back on the API key overriding PR while we think of a better way to do it? Also happy to jump on a call if it helps to talk things through. My contact info is in my profile.

igiloh-pinecone (commented):
All in all LGTM!  
See two last comments

malexw:
Great, I hope it's good to go now, @igiloh-pinecone 

In the config file, I ended up writing something a little bit different from what you recommended. When I looked at it with fresh eyes today, I felt my original wording could be a little clearer. Definitely let me know if you want further changes.

igiloh-pinecone (approved):
@malexw thank you for all of your efforts and apologies for the late response!

LGTM! PR is approved.  
Note that there are quite a few conflict, due to a recent refactor we've introduced in #231. If you'd like, someone from our end can take care of the conflicts - you've already invested a lot of effort into this PR.

malexw:
@igiloh-pinecone Sounds good! I've attempted to resolve the merge myself, but I think there might be an issue around how token budgets are calculated in `chat_engine.py`. I think the calculations work when we assume that `max_tokens` is set once in the config file, but needs to be done differently if we allow for it to be overridden. I think your team probably has a better vision for how to resolve that, so I'll hand it over to you.

igiloh-pinecone:
I think the calculations work when we assume that `max_tokens` is set once in the config file, but needs to be done differently if we allow for it to be overridden.

Thanks @malexw!!  
I've taken a deeper look - the calculations you're talking about are only affected by Canopy's built-in `max_prompt_tokens` (== input tokens sent to the model), not the `max_generated_tokens` (== tokens that the model is allowed to generate).  
At runtime, users could only affect the _generated_ tokens, by sending the `max_token` parameter - which doesn't affect Canopy's behavior.

Other than that, I'm about to approve and merge this PR.  
Thank you again very much for your efforts and contribution!!

src/canopy/llm/openai.py
```diff
@@ -117,10 +119,8 @@ def chat_completion(self,
         messages = [SystemMessage(content=system_message).dict()
                     ] + [m.dict() for m in chat_history]
         try:
-            response = self._client.chat.completions.create(model=self.model_name,
```
igiloh-pinecone, line 120: @malexw I think it's better to pass the `model=` explicitly, for code readability.  
I'm doing a small change around this and pushing to your branch.

igiloh-pinecone (approved):
LGTM!

tests/unit/chat_engine/test_chat_engine.py
```diff
@@ -130,18 +130,28 @@ def test_chat(self, namespace, history_length=5, snippet_length=10):
...
-            model_params=None
+            model_params={'max_tokens': 200}
         )
 
         assert response == expected['response']
 
     @pytest.mark.parametrize("namespace", [
         None, TEST_NAMESPACE
     ])
+    @pytest.mark.parametrize("allow_model_params_override,params_override",
```
igiloh-pinecone, line 142: @malexw I've removed the copy-pasted test cases, making the existing test parameterized instead.

PR #194, opened 2023-12-02
add initial implementation of dense Cohere encoder

DosticJelena:
## Problem

Integrating Cohere embedding models.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Infrastructure change (CI configs, etc)
- [ ] Non-code change (docs, etc)
- [ ] None of the above: (explain here)

## Test Plan

Describe specific steps for validating this change.

PR #204, opened 2023-12-07
Cohere Chat API Integration

scottmx81:
## Problem

This PR is intended to add full support for Cohere to Canopy.

Describe the purpose of this change. What problem is being solved and why?

## Solution

I have created Cohere specific classes, that work with the Cohere Python SDK. I have provided a Cohere specific configuration file, which can be used to load the Cohere classes and pass model parameters to them.

This is a work in progress PR, which I am opening now to start a discussion about how we can work around certain assumptions in the codebase, that seem to be OpenAI specific. The codebase is clearly intended to be very configurable and able to work with different LLMs, but I found it challenging to map some of the abstractions to work with the Cohere SDK.

Here are some issues that I have noticed:

* The default integration with OpenAI makes a request to the chat completion API to request search queries that can be used to query Pinecone. It uses a prompt to do this. With the Cohere Chat API, there is a `search_queries_only` flag, which I believe is the preferred way to get the relevant search queries with Cohere. In order to be able to get the search queries the Cohere specific way, rather than with a prompt asking the LLM to return search queries, I created a `ClientQueryGenerator` class, which requests the search queries from the LLM class, rather than using the OpenAI specific function calling feature with `FunctionCallingQueryGenerator`. In order to keep it somewhat generic I called it `ClientQueryGenerator`, in the sense that the LLM client will get the queries.

* The Cohere Chat API has roles `USER` and `CHATBOT`. There is not a system role, but the concept of a system prompt definitely exists in classes that are not OpenAI specific. In order for this to work Cohere Chat API, I am changing the system roles to `USER`.

* The Canopy codebase has context related abstractions, but they are built around the assumption that the context will be appended as text at the end of the prompt. This may be the way OpenAI does it, but with the Cohere Chat API, it is possible to pass a `documents` parameter, which could contain the results from the Pinecode search. At the time of opening this PR, I do not have a clean solution for passing the snippets as documents to the Cohere Chat API. It effectively still works with the default behaviour of appending `Context: ` and then the JSON from `StuffingContextContent.to_text()`, but ideally we'd have a way to pass the JSON through to `CohereLLM` class, so it it could be passed to the API via the `documents` parameter. But I don't see a clean way to do that, since the context builder classes are set up to return text, and the context itself is not passed to the LLM class as an object, but rather text already appended to the system prompt. So there is not a clean way to be able to take JSON from the context class and pass it to the `self._client.chat` call.

* At the time of opening the draft PR, it is still using `OpenAITokenizer` in the `cohere.yaml` file. I will work on adding a Cohere tokenizer class.

In this PR I have pulled in the changes from PR #194. I assume that PR will get merged first and then I will rebase when done with the draft PR.

## Type of Change

- [X ] New feature (non-breaking change which adds functionality)

## Test Plan


igiloh-pinecone:
@scottmx81 You are raising a very good point about the current interface of `BaseLLM`. It was originally designed around the chat completion API of OpenAI, which was the first LLM we've implemented.  

In the coming week, we will do a refactor of `BaseLLM` and change the `chat()` interface to take `messages()`, `system_prompt` and `context` arguments separately. The `PromptBuilder` logic that currently happens at the `ChatEngine` will move into `OpenAILLM`, and will be just one private case of how to send these 3 separate inputs to the LLM. This will allow you to utilize Cohere's API much more natively.

As for the QueryGenerator - feel free to just call it `CohereQueryGenerator`. We're considering another, separate refactor in which we'll incorporate the Query generation capabilities into the LLM classes themselves.

igiloh-pinecone:
Hi @scottmx81,

Thanks again for raising our awareness to the incompatibility of the current `BaseLLM` interface with LLMs other than OpenAI's.  
In the coming few days we will refactor `BaseLLM`'s `chat()` interface, so it will take separate `messages`, `system_prompt`, and `context` arguments. The `PromptBuilder` logic which is now part of `ChatEngine` would move into `OpenAILLM`. This way each `LLM` derived class would define its own way for using these inputs, namely you could full utilize Cohere's `chat()` API.

As for `QueryGenerator` - feel free to just call it `CohereQueryGenerator`. We're considering to also move this functionality into the individual `LLM` class (though that will probably be a separate refactor).

On that note - we're deliberating whether all classes that derive from `BaseLLM` should implement `enforce_function_call()` (not just for query generation, but also for other general purpose functions).  
Does Cohere have a similar concept?

scottmx81:
@igiloh-pinecone Thank you for considering our PR and being open to refactoring.

As far as I am aware, Cohere does not have any sort of functional calling feature. With Cohere, it is possible to call custom connectors for RAG which can be configured through the Cohere website, but there is no function calling feature as currently exists with OpenAI, where you can provide a JSON schema for function calls and have the LLM request one of those functions be called. Therefore I don't think the `enforce_function_call()` method can be implemented for search or general purpose functions with Cohere.

I have rebased the branch as requested.

Regarding the CohereTokenizer class, Cohere is recommending that we use the Cohere API for tokenization. This is different from the Open AI code, which is using a Python library to do the tokenization. There are Python files on Hugging Face which will do tokenization for specific Cohere models, but for a low-maintenance solution it is being recommended that we use the API, which is kept up to date and should work for all Cohere models.

As a result of that, I have added tests for `CohereTokenizer` under system tests, instead of unit tests, since it depends on the API and having valid API credentials. The tests that I added are largely copy/pasted from the Open AI and base tokenizer tests. I don't love copy/pasting but I don't know that it is appropriate for the "system" tests to be dependent on the unit test files. Please let me know if what I did is ok or if you think `TestCohereTokenizer` in the system tests should be inheriting from `BaseTestTokenizer` in the unit tests.

acatav:
Hi @scottmx81 thanks for your contributions! see PR #231 which separates the params for the LLM chat completion call. It will probably merge in the coming days, but you already can go ahead and adjust your PR with it

scottmx81:
I have begun updating the PR to work with the `system_prompt` and `context` arguments that are being passed to the LLM class now. Changing it to use the `system_prompt` was pretty straight forward, and I have it being passed as `preamble_override` to the Cohere Chat API.

I have found that working with the `context` argument is still less than ideal, and I don't love the solution that I currently have to read from it and pass as `documents`.  I was trying to explain to @giannis2two today the problem I am currently facing, and will describe it here for the benefit of others.

The issue right now is that the `Context` model only has a `Context.to_text` method, and there is no other common interface to pull out the documents or snippets. The code as I have it right now does get the snippets and pass them as documents, but it does so in a way that is not very clean. Because there is no interface to get the snippets, I have it attempting to parse the JSON from the `to_text()` method, and if it is successful, it loops over the items in the context, and if it sees a `snippet` key, it appends that snippet object to the `documents` list. This works but it is not a clean solution.

The reason is because the context engine can be configured, and there is no common interface that the different engines must implement to return the snippets. The only method to get at them from `context` is `to_text()`, or going through `context.content`, and pulling out snippets. But because there is no common interface, and the context engine is configurable, it's not a great idea to be looking for snippets this way. I think it would be nice if there was a way to get those snippet objects through a common interface.

For the benefit of @giannis2two and anyone else interested in this issue but has less familiarity with the codebase, I'll show an example of what could be returned from `context.to_text()`:

```
(Pdb) context.to_text()
'[{"query": "Python function result caching based on arguments", "snippets": [{"source": "https://docs.python.org/3/library/statistics.html", "text": "</pre></div>\\n</div>\\n<p>The final prediction goes to the largest posterior. This is known as the\\n<a class=\\"reference external\\" href=\\"https://en.wikipedia.org/wiki/Maximum_a_posteriori_estimation\\">maximum a posteriori</a> or MAP:</p>"}, {"source": "https://docs.python.org/3/library/zoneinfo.html", "text": "<a class=\\"reference internal\\" href=\\"#zoneinfo-data-configuration\\"><span class=\\"std std-ref\\">Configuring the data sources</span></a>.</p>\\n</dd></dl>"}, {"source": "https://docs.python.org/3/library/sysconfig.html", "text": "<span class=\\"go\\">Current installation scheme: \\"posix_prefix\\"</span>"}, {"source": "https://docs.python.org/3/library/pickle.html", "text": "<p><span class=\\"versionmodified changed\\">Changed in version 3.8: </span>The default protocol is 4.</p>\\n</div>\\n</dd></dl>"}, {"source": "https://docs.python.org/3/library/graphlib.html", "text": "<a class=\\"reference internal\\" href=\\"#graphlib.TopologicalSorter.prepare\\" title=\\"graphlib.TopologicalSorter.prepare\\"><code class=\\"xref py py-meth docutils literal notranslate\\"><span class=\\"pre\\">prepare()</span></code></a> previously.</p>\\n</dd></dl>"}]}]'
```

What it is doing in the current version of this PR is parsing this JSON, then looking for the `snippet` array, and then including those as the documents. It works OK. But this is tying the `llm/cohere.py` module to internal details of `StuffingContextContent` class. In theory, users are not required to use `StuffingContextBuilder`, since it can be configured to something else in the YAML config file, although currently `StuffingContextBuilder` is the only context builder class that exists. But since it is set up in a way to be configurable by the user to something else, it's not right for the `llm/cohere.py` code to be assuming that `StuffingContextBuilder` is being used. In my opinion it would be ideal for there to be another way to get at the snippets. For example, if there were a `snippets` property on the `canopy.models.data_models.Context` model itself.

When I go through `context.content`, instead of parsing the JSON from `.to_text()`, I can see the models `ContextQueryResult` and `ContextSnippet`. At first glance it would appear that these define a common interface, but upon further investigation I see these are specific to `canopy.context_engine.context_engine.stuffing`, as such they aren't really common models/interfaces that other (future) context engines would use, and therefore it's not really appropriate for code in `llm/cohere.py` to be specifically looking for them. I was thinking that it might be appropriate for those to be in `canopy.context_engine.context_engine.base` or `canopy.models.data_models`, so it wouldn't be as "wrong" for the `CohereLLM` class to be looking for them (such as with an isinstance check).

```
(Pdb) context
Context(content=StuffingContextContent(__root__=[ContextQueryResult(query='Python function result caching based on arguments', snippets=[ContextSnippet(source='https://docs.python.org/3/library/statistics.html', text='</pre></div>\n</div>\n<p>The final prediction goes to the largest posterior. This is known as the\n<a class="reference external" href="https://en.wikipedia.org/wiki/Maximum_a_posteriori_estimation">maximum a posteriori</a> or MAP:</p>'), ContextSnippet(source='https://docs.python.org/3/library/zoneinfo.html', text='<a class="reference internal" href="#zoneinfo-data-configuration"><span class="std std-ref">Configuring the data sources</span></a>.</p>\n</dd></dl>'), ContextSnippet(source='https://docs.python.org/3/library/sysconfig.html', text='<span class="go">Current installation scheme: "posix_prefix"</span>'), ContextSnippet(source='https://docs.python.org/3/library/pickle.html', text='<p><span class="versionmodified changed">Changed in version 3.8: </span>The default protocol is 4.</p>\n</div>\n</dd></dl>'), ContextSnippet(source='https://docs.python.org/3/library/graphlib.html', text='<a class="reference internal" href="#graphlib.TopologicalSorter.prepare" title="graphlib.TopologicalSorter.prepare"><code class="xref py py-meth docutils literal notranslate"><span class="pre">prepare()</span></code></a> previously.</p>\n</dd></dl>')])]), num_tokens=384, debug_info={'num_docs': 5})
```

In other words, it seems like I could just loop over `context.content.__root__` and get the snippets from the `ContextQueryResult` objects, but the point I'm trying to explain is that there's no guarantee that will work with other context builders, since the context class is configurable in the YAML config file, and it might not be `StuffingContextBuilder`. So it would be nice if there was a common interface to get them instead of accessing "private" data in the stuffing context classes.

I'd appreciate any direction from the Canopy team about whether it's OK to get the snippets the way I have it, or if you would agree that a common interface to get snippets would be a good idea.

igiloh-pinecone:
> I have found that working with the `context` argument is still less than ideal, and I don't love the solution that I currently have to read from it and pass as `documents`. I was trying to explain to @giannis2two today the problem I am currently facing, and will describe it here for the benefit of others.

Hi @scottmx81, thank you very much for your feedback!!  
You are absolutely correct that at the time being, the only "guaranteed" interface for `Context` is its `.to_text()` method. This was by design - under the assumption that all LLMs eventually represent the context as a single contiguous array of tokens. I'm guessing Cohere's LLM implementation eventually does the same - it simply happens on the server side, using dedicated document delimiters etc.  

It might very well be that in the future once we start incorporating more advanced `ContextBuilder`s, we would need to better define the interface of `Context`. But at this current point in time this is still undecided.  
For the time being, I think we are perfectly happy with a solution that does something like
```python
if isinstance(context.content, StuffingContextContent):
    # Use the structure of StuffingContextContent to build documents
    documents = ...
else:
    # Assume unknown JSON structure
    documents = json.loads(...)
```
Or even a more extreme
```python
if isinstance(context.content, StuffingContextContent):
    # Use the structure of StuffingContextContent to build documents
    documents = ...
else:
    raise NotImplementedError("Cohere LLM is currently supported only with....")
```

Does that unblock you? Would you like to further discuss together?

scottmx81:
@igiloh-pinecone Yes, this works. I will update the PR today. Thanks.

igiloh-pinecone (changes requested):
@scottmx81, thank you very much for your contribution!! There was clearly a lot of effort dedicated to this PR.  
Apologies for the delayed review!

Please see a few comments and suggestions. Feel free to discuss or offer alternatives, nothing is set in stone.

tests/system/query_generator/test_cohere_query_generator.py
```diff
@@ -0,0 +1,30 @@
+import pytest
+
+from canopy.chat_engine.query_generator.cohere import CohereQueryGenerator
+from canopy.models.data_models import MessageBase, Role
+
+
+@pytest.fixture
+def messages():
```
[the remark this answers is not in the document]
scottmx81: @igiloh-pinecone I changed it to use the fixture from `conftest.py` in the other place you mentioned it, but I ran into an issue with doing it here. The problem is that the last message from the fixture is "Just checking in. Be concise.", but because this message does not have a question from the user, the Cohere Chat API with the `generate_search_queries=True` parameter does not return any recommended search queries.

I guess I could use that fixture and then append a real question to the messages in the test.

src/canopy/tokenizer/cohere.py
```diff
@@ -0,0 +1,91 @@
+import cohere  # type: ignore
+
+from typing import List, Optional
+from .base import BaseTokenizer
+from ..models.data_models import Messages
+
+
+class CohereTokenizer(BaseTokenizer):
```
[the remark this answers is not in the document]
scottmx81: @igiloh-pinecone Yes, I wasn't entirely convinced on using the API for tokenizing the text either, however I asked the team at Cohere about it, and they seemed to think it was OK to use the API, so I went ahead with implementing it that way. They may not have had the full context of the problem however, in that it's only being used for token counting.

When I had asked them before, they did point me to some Cohere tokenizers on HF, but they were specific to certain models, and said the latest ones were available via the API. But maybe it's good enough to use one of the Python ones for an approximate token count only, instead of the performance implications from hitting the API.

I'd suggest leaving the Cohere Tokenizer as I've done it with the API as an option for the user, but not making it the default. That way the user could decide what is more important to them - more accurate token counting vs performance. Maybe it could be renamed from `CohereTokenizer` to `CohereAPITokenizer`.

https://huggingface.co/Cohere/Command-nightly

I'll raise the issue with Cohere today and see what they suggest also.

src/canopy/llm/cohere.py
```diff
@@ -0,0 +1,284 @@
...
+            base_url: The base URL to use for the Cohere API. Defaults to None (uses the "CO_API_URL" environment variable if set, otherwise use default Cohere API URL).
+            **kwargs: Generation default parameters to use for each request. See https://platform.openai.com/docs/api-reference/chat/create
+                    For example, you can set the temperature, p, etc
+                    These params can be overridden by passing a `model_params` argument to the `chat_completion` methods.
+        """  # noqa: E501
+        super().__init__(model_name)
+        self._client = cohere.Client(api_key, api_url=base_url)
+        self.default_model_params = kwargs
+
+    def chat_completion(self,
```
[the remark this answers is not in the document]
scottmx81: @igiloh-pinecone I have checked the source code for the Cohere Python client, and I have confirmed that the Cohere client does have its own retry functionality in the current 4.x version. However, in the branches for the v5.0 alpha releases, that parameter is no longer there. I have asked for clarification on whether we will be able to count on the retry functionality continuing to exist in the Cohere client.

src/canopy/tokenizer/cohere.py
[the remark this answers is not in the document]
igiloh-pinecone: A note to selves (unrelated to this PR) is to try to optimize \ reduce the call to tokenizers, and maybe even rename them `TokenCounter` which better suits their current functionality

tests/system/query_generator/test_cohere_query_generator.py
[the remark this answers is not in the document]
igiloh-pinecone: If it's only one test case - I guess appending a question inside the test func sounds right.  
Up to you.

src/canopy/tokenizer/cohere.py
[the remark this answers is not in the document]
scottmx81: @igiloh-pinecone they have shown me how I can do it with HuggingFace's `tokenizer` library. I am going to work on changing it now.

igiloh-pinecone:
@scottmx81 thank you again for all of your efforts and dedication on this PR!!!  
Your continuing effort hasn't gone unnoticed, and we truly appreciate your contribution.

While giving the PR one last glance before approving it, I noticed some possible improvement for Tokenizer testing. To avoid another round if reviews and fixes - I simply opened a [change suggestion](https://github.com/scottmx81/canopy/pull/1). Let me know what you think.

Looking forward already for merging this PR!

src/canopy/llm/cohere.py
```diff
@@ -0,0 +1,284 @@
...
+        Usage:
+            >>> from canopy.llm import OpenAILLM
+            >>> from canopy.models.data_models import UserMessage
+            >>> llm = CohereLLM()
+            >>> messages = [UserMessage(content="Hello! How are you?")]
+            >>> result = llm.chat_completion(messages)
+            >>> print(result.choices[0].message.content)
+            "I'm good, how are you?"
+        """  # noqa: E501
+        model_params_dict: Dict[str, Any] = deepcopy(self.default_model_params)
```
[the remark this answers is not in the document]
igiloh-pinecone: @scottmx81 On second thought - let get this PR merged already! 😃  
We can always add this feature separately later

src/canopy/llm/cohere.py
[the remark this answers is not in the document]
aalpat4: Retry is indeed supported, and will also be in stable v5.x releases! @scottmx81 all good to use the 4.x retry functionality

config/cohere.yaml
```diff
@@ -0,0 +1,73 @@
...
+  params:
+    system_prompt: *system_prompt
+
+  # -------------------------------------------------------------------------------------------------------------
+  # LLM configuration
+  # -------------------------------------------------------------------------------------------------------------
+  llm: &llm
+    type: CohereLLM
+    params:
+      model_name: command-nightly
```
aalpat4: let's use `command` model here

src/canopy/chat_engine/query_generator/cohere.py
```diff
@@ -0,0 +1,42 @@
...
+        messages = self._history_pruner.build(chat_history=messages,
+                                              max_tokens=max_prompt_tokens)
+        llm = cast(CohereLLM, self._llm)
+        queries = llm.generate_search_queries(messages)
+        return [Query(text=query) for query in queries]
+
+    async def agenerate(self,
+                        messages: Messages,
+                        max_prompt_tokens: int) -> List[Query]:
+        raise NotImplementedError
```
aalpat4, line 42: same comment as below, but we do have an Async client in the Cohere Python SDK

src/canopy/llm/cohere.py
```diff
@@ -0,0 +1,325 @@
...
+
+        if not messages:
+            raise cohere.error.CohereAPIError("No message provided")
+
+        response = self._client.chat(
+            model=self.model_name,
+            message=messages.pop()['message'],
+            chat_history=messages,
+            documents=self.generate_documents_from_context(context),
+            preamble_override=system_prompt,
```
aalpat4, line 107: By default, this should actually be empty since Cohere handles the system prompt for you as long as the documents parameter is used. So perhaps we can replace "system_prompt" with "preamble_override" as a Cohere-only supported parameter?

src/canopy/tokenizer/cohere.py
```diff
@@ -0,0 +1,185 @@
+from typing import List, Optional
+
+import cohere
+from tokenizers import Tokenizer as HfTokenizer
+
+from .base import BaseTokenizer
+from ..models.data_models import Messages
+
+
+class CohereHFTokenizer(BaseTokenizer):
```
aalpat4, line 10: nice 🙌

aalpat4 (changes requested):
This is great - really excited for this! Left a couple of comments on how I think we can better align with the Cohere API (e.g. async support, default preamble). 

Please reach out if there are any questions

src/canopy/llm/cohere.py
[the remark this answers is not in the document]
scottmx81: @igiloh-pinecone I didn't see this last comment until now, and I have already implemented a `_convert_model_params` function as requested. The change has been pushed.

config/cohere.yaml
```diff
@@ -0,0 +1,73 @@
...
+      # You can add any additional parameters which are supported by the Cohere Co.Chat API. The values set
+      # here will be used in every Co.Chat API call. For example:
+      prompt_truncation: "AUTO"
+      # Specifying connectors is contrary to Canopy's purpose of searching the Pinecone knowledge base only,
+      # but technically can still be passed like this:
+      # connectors:
+      #  - "web-search"
+      citation_quality: "accurate"
+      temperature: 0.85
```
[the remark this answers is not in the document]
scottmx81: Yes, good catch @aalpat4 . I added them as an example since the main `config.yaml` file shows how to pass custom model params also, but they are commented out by default. I intended for them to be shown as an example only in the `cohere.yaml` config file, but at some point during local dev I uncommented them so that I could see the custom params being passed in during local dev. I will comment them out, but leave them as comments as guidance for users to see how custom model params can be passed in this way.

src/canopy/chat_engine/query_generator/cohere.py
scottmx81: @aalpat4 yes, you are right again. I did notice at some point that there is an async Cohere client. However I also notice that the other LLM backends in Canopy do not yet support the async versions of the functions, and two of the three other query generators do not either, including the default one. I think it would be best to try to get this PR merged in without the async functions first, and do the async implementation in another PR.

src/canopy/llm/cohere.py
```diff
@@ -210,6 +212,44 @@ async def agenerate_queries(self,
...
+        """
+        Convert Open AI model params to Cohere equivalents.
+
+        Args:
+            openai_model_params: model params passed from client to Canopy API in OpenAI format.
+
+        Returns:
+            Model params used with Cohere Chat API.
+        """  # noqa: E501
+        common_params = [
```
giannis2two, line 227: As discussed, larger surface area can be covered here with https://cohere-sdk.readthedocs.io/en/latest/cohere.html#cohere.client.Client.chat params

src/canopy/llm/cohere.py
```diff
@@ -0,0 +1,327 @@
...
+        Usage:
+            >>> from canopy.llm import OpenAILLM
+            >>> from canopy.models.data_models import UserMessage
+            >>> llm = CohereLLM()
+            >>> messages = [UserMessage(content="Hello! How are you?")]
+            >>> result = llm.chat_completion(messages)
+            >>> print(result.choices[0].message.content)
+            "I'm good, how are you?"
+        """  # noqa: E501
+        model_params_dict: Dict[str, Any] = deepcopy(self.default_model_params)
```
giannis2two, line 88: May be incorrect here if already tried, but don't think a `deepcopy` is needed here

src/canopy/llm/cohere.py
```diff
@@ -0,0 +1,327 @@
...
+
+        # Scale is -2.0 to 2.0 with OpenAI, but -1.0 to 1.0 with Cohere.
+        if presence_penalty := converted_model_params.get("presence_penalty"):
+            converted_model_params = presence_penalty * 0.5
+
+        unrecognized_keys = set(openai_model_params.keys())
+        default_keys = set(self.default_model_params.keys())
+
+        if unrecognized_keys.difference(default_keys) and not self.ignore_unrecognized_params:
+            raise NotImplementedError(
```
giannis2two, line 249: Could maybe raise a warning here instead of failing, unless otherwise discussed in this PR already

config/cohere.yaml
```diff
@@ -0,0 +1,73 @@
+# ==================================================================
+#            Configuration file for Canopy Server with Cohere.
+# ==================================================================
+
+# ---------------------------------------------------------------------------------
+system_prompt: &system_prompt |
+  Use the documents to answer the user question at the next messages. The documents are retrieved from a knowledge
```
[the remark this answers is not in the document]
giannis2two: +1 here. When specifying `documents` in the SDK, our RAG preamble gets used here to generate an appropriate response with the given documents (somewhat similar to how `query_builder_prompt` is not used in `cohere.yaml` due to our `search_queries` boolean). In Cohere case, `system_prompt` should be an optional knob for someone that wants to experiment with a custom RAG preamble.

The only issue here I believe is that there is a `DEFAULT_SYSTEM_PROMPT` in `chat_engine.py` that gets used if this is omitted? Can work something out here.

config/cohere.yaml
```diff
@@ -0,0 +1,75 @@
...
+    knowledge_base:
+      # --------------------------------------------------------------------------
+      # Configuration for the RecordEncoder subcomponent of the knowledge base.
+      # --------------------------------------------------------------------------
+      record_encoder:
+        type: CohereRecordEncoder
+        params:
+          model_name:                   # The name of the model to use for encoding
+            "embed-english-v3.0"
+          batch_size: 100               # The number of document chunks to encode in each call to the encoding model
```
giannis2two, line 75: An issue I noticed here is that our embed v3 models here uses 1024 dimension by default. Unless otherwise specified, Canopy creates a Pinecone index with 1536 dimensions, the OAI default. Passing the following specifictations below (similar to `config.py`) helps this issue

```
create_index_params:
  # -------------------------------------------------------------------------------------------
  # Initialization parameters to be passed to create a canopy index. These parameters will
  # be used when running "canopy new".
  # -------------------------------------------------------------------------------------------
  metric: cosine
  dimension: 1024
  spec:
    serverless:
      cloud: aws
      region: us-west-2
```

src/canopy/llm/cohere.py
```diff
@@ -0,0 +1,327 @@
...
+            raise cohere.error.CohereAPIError("No message provided")
+
+        response = self._client.chat(
+            model=self.model_name,
+            message=messages.pop()['message'],
+            chat_history=messages,
+            documents=self.generate_documents_from_context(context),
+            preamble_override=system_prompt,
+            stream=stream,
+            connectors=[
```
giannis2two, line 109: This is a nice and slick addition but maybe not directly relevant to Canopy, somewhat defeating the purpose. Okay with me to leave here, but can let @igiloh-pinecone to chime in as well.

giannis2two (commented):
Leaving a couple comments below. Thank you +100 Scott and Illai. Having an internal discussion about a leftover idea here otherwise looking great!

src/canopy/chat_engine/query_generator/cohere.py
igiloh-pinecone: @scottmx81 @aalpat4 we are planning to implement full Async support across all system parts (embedding, LLM etc) very soon. But as @scottmx81 said, right now it's not a requirement for adding Cohere LLM support.  
We can work on this together as a separate effort later

config/cohere.yaml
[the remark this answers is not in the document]
igiloh-pinecone: Canopy allows users to change the system prompt, even in the RAG case, so they can add their own tweaks like avoiding specific topics, using specific tones etc.  
Unless there is a very strong incentive to do otherwise, we would like to keep this behavior.

@aalpat4 \ @giannis2two can we simply put the default Cohere RAG prompt here in the config and leave it transparent for users to change?

Your suggestion is a bit problematic, as it breaks Canopy's design principles. Note that this `system_prompt` key configures the `ChatEngine` itself, not the underlying `LLM` class.  
The `CohereLLM` class is considered a stand alone, **stateless** LLM wrapper, which takes   `system_prompt` (==preamble) and `context` (==`documents`) inputs **per `chat()` call**. Theoretically the same `LLM` instance might be used to do RAG in a one `chat()` call, and a completely different task in the next call (with a different prompt, without documents etc).

src/canopy/llm/cohere.py
```diff
@@ -0,0 +1,325 @@
...
+        messages: List[Dict[str, Any]] = self._map_messages(chat_history)
+
+        if not messages:
+            raise cohere.error.CohereAPIError("No message provided")
+
+        response = self._client.chat(
+            model=self.model_name,
+            message=messages.pop()['message'],
+            chat_history=messages,
+            documents=self.generate_documents_from_context(context),
```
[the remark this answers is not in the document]
igiloh-pinecone: As I wrote above about the `system_prompt` config param - in Canopy the `LLM` class is assumed to be standalone and stateless.  
The thing wrapping\using the LLM (like `ChatEngine` in the case of RAG task) is responsible for setting the fields it passes to the underlying LLM. This is currently agnostic to the type of underlying LLM (e.g. `OpenAILLM` vs `CohereLLM`).  

The `CohereLLM` class itself should make any assumptions that it is always being used for RAG. In the future we might add some other use cases\ pipelines, similar to our current `ChatEngine`, which uses the `LLM` in some other manner

src/canopy/llm/cohere.py
igiloh-pinecone: See my comments above. The `LLM` class shouldn't assume the use case it is being used for (e.g. RAG)

src/canopy/llm/cohere.py
igiloh-pinecone: My only concern with leaving it a warning is that we slightly fail in case someone has accidentally misspelled one of the supported parameters. They will get some undesired behavior, without even knowing (unless they check the logs).

config/cohere.yaml
igiloh-pinecone: @giannis2two, we're actually just in the process of removing `dimension` from `create_index_params` altogether.  
It was always an optional param, as Canopy tries to infer the dimension from the configured `record_encoder `. Now we simply decided it's not even optional anymore... 😄

tests/unit/tokenizer/test_cohere_hf_tokenizer.py
igiloh-pinecone, line 1: @scottmx81 please see my [suggested changes](https://github.com/scottmx81/canopy/pull/1) for Tokenizer testing.  
Feel free to just merge agree with the suggestion (it will merge it into this PR)

config/cohere.yaml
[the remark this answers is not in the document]
giannis2two: Understood thanks for the details here @igiloh-pinecone. Let us check internally regarding `can we simply put Cohere's RAG prompt here`. Can think of another potential solution here as well. Will come back here.

src/canopy/llm/cohere.py
scottmx81: @giannis2two I would suggest we keep the `deepcopy`. I don't think it's an issue. Also, that's how they do it in the `OpenAILLM.chat_completion` method - with the deep copy.

The reason is because of how the model params are merged. One set is passed to the `CohereLLM.__init__` method (which come from the YAML config file and apply to all chat completion calls), and then there are separate model params passed to the `CohereLLM.chat_completion` method. Those two sets of model params get merged, and if it didn't do the deep copy here, there would be risk of updating values in the `self.self.default_model_params` dict with values from `model_params` passed to the chat completion method. There's certainly other ways of merging the values of the two dicts together without using `deepcopy`, but that is the method they chose to merge the default model params and the ones passed on each chat completion call in the base `OpenAILLM.chat_completion` method, and I'm doing it the same way here.

src/canopy/llm/cohere.py
scottmx81: @giannis2two I have updated this method to handle a few of the parameters that were previously undocumented.

I have added `logit_bias` to the `common_params` list, and I have added `top_p` and `user` and mapped them to the Cohere Co.Chat API's `p` and `user_name` parameters.

config/cohere.yaml
giannis2two: >  as Canopy anyway tries to infer the dimension from the configured record_encoder

@igiloh-pinecone please disregard. Seems like a local configuration issue when I first created the index through the CLI. Thanks for the details!

src/canopy/llm/cohere.py
```diff
@@ -0,0 +1,325 @@
...
+                               system_prompt: str,
+                               chat_history: Messages,
+                               context: Optional[Context] = None,
+                               *,
+                               stream: bool = False,
+                               max_generated_tokens: Optional[int] = None,
+                               model_params: Optional[dict] = None,
+                               ) -> Union[ChatResponse,
+                                          Iterable[StreamingChatChunk]]:
+        raise NotImplementedError("Cohere LLM doesn't support async chat completion")
```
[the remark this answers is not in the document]
scottmx81: As discussed in the other comments, async support can be added later.

src/canopy/llm/cohere.py
```diff
@@ -0,0 +1,325 @@
...
+
+    async def agenerate_queries(self,
+                                messages: Messages,
+                                *,
+                                max_generated_tokens: Optional[int] = None,
+                                model_params: Optional[dict] = None,
+                                ) -> List[Query]:
+        raise NotImplementedError("Cohere LLM doesn't support async query generation")
```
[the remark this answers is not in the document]
scottmx81: As mentioned in the other comments, we can add async support after, in another PR.

config/cohere.yaml
scottmx81: It has been changed to `command`.

tests/unit/tokenizer/test_cohere_hf_tokenizer.py
scottmx81: Thank you for your suggested changes. I have merged your PR.

src/canopy/llm/cohere.py
```diff
@@ -0,0 +1,284 @@
...
+        Args:
+            model_name: The name of the model to use. See https://docs.cohere.com/docs/models
+            api_key: Your Cohere API key. Defaults to None (uses the "CO_API_KEY" environment variable).
+            base_url: The base URL to use for the Cohere API. Defaults to None (uses the "CO_API_URL" environment variable if set, otherwise use default Cohere API URL).
+            **kwargs: Generation default parameters to use for each request. See https://platform.openai.com/docs/api-reference/chat/create
+                    For example, you can set the temperature, p, etc
+                    These params can be overridden by passing a `model_params` argument to the `chat_completion` methods.
+        """  # noqa: E501
+        super().__init__(model_name)
+        self._client = cohere.Client(api_key, api_url=base_url)
```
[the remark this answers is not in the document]
scottmx81: @igiloh-pinecone I have added the `try`-`except` here as requested. I did the test a little bit different from the equivalent for OpenAI - using the pytest `monkeypatch` fixture to set the env var instead of a custom `bad_api_key` fixture.  I think this way is more elegant but I can change it to do it the other way for consistency if you'd like.

src/canopy/llm/cohere.py
```diff
@@ -0,0 +1,284 @@
...
+        model_params_dict.update(
+            model_params or {}
+        )
+        connectors = model_params_dict.pop('connectors', None)
+        messages: List[Dict[str, Any]] = self.map_messages(chat_history)
+
+        if not messages:
+            raise cohere.error.CohereAPIError("No message provided")
+
+        response = self._client.chat(
```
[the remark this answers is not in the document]
scottmx81: @igiloh-pinecone It has been confirmed that the Cohere client handles retries for transient errors itself.

For the other errors, I have added the `try`-`except` clause as requested, to re-raise the `CohereAPIError` exceptions as `RuntimeError` with a message based on the `OpenAILLM` equivalent.

Note that I have added this exception handling in two separate places - at instantiation it catches `CohereError`, and around the `self._client.chat` call it catches `CohereAPIError`. I checked the source code for the Cohere client and `CohereError` for the API key is the only error it raises from the `__init__` method.

giannis2two (approved):
Looking good from our side.

src/canopy/llm/cohere.py
```diff
@@ -95,22 +105,33 @@ def chat_completion(self,
...
-            stream=stream,
-            connectors=[
-                {"id": connector} for connector in connectors
-            ] if connectors else None,
-            **model_params_dict
-        )
+            raise RuntimeError("No message provided")
+
+        if system_prompt:
+            messages = self._prepend_system_prompt_to_messages(system_prompt, messages)
```
igiloh-pinecone, line 114: @scottmx81 @giannis2two I guess we can live with this solution for now.

I can think of some theoretical case when someone is using the `CohereLLM` class with a single instruction, without `context`. In this case sending his prompt as user message instead of `preamble_override` might  deteriorate the result, right?  
But that's totally an edge case for now, Canopy is mostly focused on RAG.

tests/system/tokenizer/test_cohere_api_tokenizer.py
```diff
@@ -1,91 +1,55 @@
...
         assert tokenizer.messages_token_count([]) == 3
 
-    # endregion
-
-    # region special tokens
-
     @staticmethod
     def test_special_tokens_to_natural_text(tokenizer):
-        tokens = tokenizer.tokenize("<|endoftext|>")
+        tokens = tokenizer.tokenize("<BOS_TOKEN>")
```
igiloh-pinecone, line 50: @scottmx81 Please note that this test still fails - for both `test_cohere_api_tokenizer` and `test_cohere_hf_tokenizer`.  
The expected behavior is that the string `'<BOS_TOKEN>'` is treated as a string, not an actual token (as it may naturally occur in the user's text).  

Can you please work with the Cohere people to see if there's a way to configure "ignore token-like strings"?

igiloh-pinecone (approved):
@scottmx81 all in all LGTM!

See a few minor comments, and please note one system tests that fails and needs to be handled (or skipped if there is no easy solution).  
Other than that PR is approved

src/canopy/llm/cohere.py
giannis2two: Indeed this is valid. The delta of `preamble_override` vs "prompt as user message" should not have big impact for non-RAG behaviors. What we are really trying to optimize with this here is our RAG preamble that has noticeable impact when it is replaced.

src/canopy/llm/cohere.py
```diff
@@ -227,25 +248,37 @@ def _convert_model_params(self, openai_model_params: dict) -> dict:
...
+
         converted_model_params = {}
 
         for param in common_params:
             if param in openai_model_params:
                 converted_model_params[param] = openai_model_params.pop(param)
 
+        for param, equivalent in equivalent_params.items():
+            if param in openai_model_params:
+                converted_model_params[equivalent] = openai_model_params.pop(param)
```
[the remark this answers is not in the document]
scottmx81: @igiloh-pinecone The problem with the suggested change is that it's not allowed to pop items from a dict while looping over it. Even with `for param in openai_model_params.keys():` it's still a runtime error. That is why I have slightly modified the suggested change to cast it to a list, so that we are iterating over the list, not the dict. It's effectively the same, though.

src/canopy/llm/cohere.py
```diff
@@ -227,25 +248,37 @@ def _convert_model_params(self, openai_model_params: dict) -> dict:
         common_params = [
```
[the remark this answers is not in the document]
scottmx81: Done and done.

tests/system/tokenizer/test_cohere_api_tokenizer.py
scottmx81: @igiloh-pinecone I have looked into this today and raised the issue, and it turns out that Cohere does not support having special token symbols in the prompt. This is the expected behaviour from the Cohere tokenizers right now. These special token symbols in the prompts will effectively be dropped by the tokenizers.

I know that there are edge cases where a user may want to do this, such as if the user were to ask the chatbot what that token symbol means. This is an edge case that Cohere does not support right now.

Therefore I think it is best to remove this test. To the best of my knowledge, there is no escape function or "ignore token-like strings" flag. The count being returned is the real and expected count, at least as far as the Cohere tokenizers are concerned.

src/canopy/llm/cohere.py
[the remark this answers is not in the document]
igiloh-pinecone: Yeah, you're absolutely right, I missed that.

tests/system/tokenizer/test_cohere_api_tokenizer.py
igiloh-pinecone: Understood. I agree this is an edge case, we can skip the test or simply delete it.

src/canopy/llm/cohere.py
```diff
@@ -0,0 +1,390 @@
+import time
+from copy import deepcopy
+from typing import Union, Iterable, Optional, Any, Dict, List
+
+import cohere
```
acatav, line 5: cohere is an optional dependency, and we import it for any config in `llm/__init__.py` so please follow the try catch we apply in this cases. For example see `src/canopy/tokenizer/llama.py`

acatav: Please make sure we are not directly import cohere in any place around the project

config/cohere.yaml
```diff
@@ -0,0 +1,75 @@
...
+    type: CohereQueryGenerator
+    params: {}
+    llm:
+      <<: *llm
+
+
+  # -------------------------------------------------------------------------------------------------------------
+  # ContextEngine configuration
+  # -------------------------------------------------------------------------------------------------------------
+  context_engine:
```
acatav, line 62: We can do it in another PR but I think that it make sense having the cohere reranker part of the cohere default config file

config/cohere.yaml
```diff
@@ -0,0 +1,75 @@
...
+# Chat engine configuration
+# -------------------------------------------------------------------------------------------------------------
+chat_engine:
+  params:
+    system_prompt: *system_prompt
+
+  # -------------------------------------------------------------------------------------------------------------
+  # LLM configuration
+  # -------------------------------------------------------------------------------------------------------------
+  llm: &llm
```
acatav, line 33: Can we make sure that the value for max tokens make sense here for Cohere API? it also seems like it make sense to have the default a huge number, and use `prompt_truncation=True` to let Cohere manage the truncation

acatav (changes requested):
LGTM, great work @scottmx81 !!

See my small comments

src/canopy/llm/cohere.py
igiloh-pinecone: Good catch!! 

@scottmx81 could you please make that change?

config/cohere.yaml
giannis2two: Using `max_tokens=None` the model will generate a completion until EOS. Aside from that, no need for a default number to be suggested; up to user's discretion imo. `prompt_truncation = AUTO` is also the API default so would happen automatically.

acatav (approved):

igiloh-pinecone:
@scottmx81 @giannis2two  
I just tested this branch manually, and I believe there is a bug in `CohereQueryGenerator`.  
It seems that when `search_queries_only=True`, the model completely ignore the `message=` argument, and only looks in the `chat_history`. So the for first message in every new chat conversation - no queries are generated (an empty list is returned). 

I have re-tested this several times, and it seems to be a consistent behavior. Can you please check before we merge?

giannis2two:
> @scottmx81 @giannis2two I just tested this branch manually, and I believe there is a bug in `CohereQueryGenerator`. It seems that when `search_queries_only=True`, the model completely ignore the `message=` argument, and only looks in the `chat_history`. So the for first message in every new chat conversation - no queries are generated (an empty list is returned).
> 
> I have re-tested this several times, and it seems to be a consistent behavior. Can you please check before we merge?

Ack. Haven't observed this behavior personally, can be due to a recent update. Will look into it with Scott.

igiloh-pinecone:
> > @scottmx81 @giannis2two I just tested this branch manually, and I believe there is a bug in `CohereQueryGenerator`. It seems that when `search_queries_only=True`, the model completely ignore the `message=` argument, and only looks in the `chat_history`. So the for first message in every new chat conversation - no queries are generated (an empty list is returned).
> > I have re-tested this several times, and it seems to be a consistent behavior. Can you please check before we merge?
> 
> Ack. Haven't observed this behavior personally, can be due to a recent update. Will look into it with Scott.

@giannis2two, sorry, never mind.  
The result I got does seem to be some sporadic behavior, which depends more on the context of my message, not a consistent behavior. My bad...

So everything SGTM. Merging!  
Thank you @scottmx81 and @giannis2two for this elaborate effort!

giannis2two:
Thank you for the help everyone. This was a loaded effort! Looking forward to working more together.

scottmx81:
@igiloh-pinecone I see the failure. I think it's because there is no `__init__.py` file in `tests/system/tokenizer/`. I will resolve the issue.

scottmx81:
@igiloh-pinecone I pushed a fix. I confirm that I can run `pytest tests/system/tokenizer/test_cohere_api_tokenizer.py` successfully in my local environment, so I think we should be good to try the merge again.

PR #208, opened 2023-12-08
Fix issue 205

### LINKED ISSUE #205: [Bug] CLI requires OpenAI API key, even if OpenAI is not being used
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

The CLI uses the `openai` client in `canopy chat` to talk with the Canopy server, regardless of the LLM actually being used (which might be non-OpenAI).  
But since `openai` client verifies the presence of an Open AI API key, this errors out, even though the configuration is correct.

We need to change [this line](https://github.com/pinecone-io/canopy/blob/main/src/canopy_cli/cli.py#L391) to pass some dummy `api_key=` value.

### Expected Behavior

`canopy chat` should work with any valid LLM config

### Steps To Reproduce

1. Unset the `OPENAI_API_KEY` env var.
2. run `canopy chat`

### Relevant log output

_No response_

### Environment

```markdown
- **OS**:
- **Language version**:
- **Canopy version**:
```

### Additional Context

_No response_

miararoy:
## Problem

dependency on OPENAI api key

## Solution

dummy setting of one in `chat` closes #205 

## Type of Change

## Test Plan

Manual

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,8 @@ This will open a chat interface in your terminal. You can ask questions and the 
 
 To compare the chat response with and without RAG use the `--no-rag` flag
 
+> **Note**: This method is only supported with OpenAI at the moment.
+
 ```bash
 canopy chat --no-rag
 ```
--- a/src/canopy_cli/cli.py
+++ b/src/canopy_cli/cli.py
@@ -382,13 +382,23 @@ def _chat(
     model,
     history,
     message,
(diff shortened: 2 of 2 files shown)


igiloh-pinecone (approved):
LGTM

PR #210, opened 2023-12-10
Add last message query generator

izellevy:
## Problem

Some LLMs do not support function calls. We need a basic query generator for this kind of APIs. 

## Solution

Add a super basic query generator that takes the last message and returns the message as is as a query.

## Type of Change

## Test Plan
Added necessary unit tests.


src/canopy/chat_engine/query_generator/last_message.py
```diff
@@ -0,0 +1,17 @@
+from typing import List
+
+from canopy.chat_engine.query_generator import QueryGenerator
+from canopy.models.data_models import Messages, Query
+
+
+class LastMessageQueryGenerator(QueryGenerator):
+
```
acatav, line 8: Please add some docstring describing the behavior and propose of this class. Also it's better to explain in the generate function that we don't use the `max_prompt_tokens` arg

src/canopy/chat_engine/query_generator/last_message.py
```diff
@@ -0,0 +1,17 @@
...
+    def generate(self,
+                 messages: Messages,
+                 max_prompt_tokens: int) -> List[Query]:
+        return [Query(text=messages[-1].content)]
```
acatav, line 12: Consider raise a value error for empty messages list with an explanation to the user

acatav (approved):

src/canopy/chat_engine/query_generator/last_message.py
izellevy: Done

src/canopy/chat_engine/query_generator/last_message.py
izellevy: Done

src/canopy/chat_engine/query_generator/last_message.py
```diff
@@ -0,0 +1,28 @@
...
+                 messages: Messages,
+                 max_prompt_tokens: int) -> List[Query]:
+        """
+            max_prompt_token is dismissed since we do not consume any token for
+            generating the queries.
+        """
+        if len(messages) == 0:
+            raise ValueError("Passed chat history does not contain any messages."
+                             " Please include at least one message in the history.")
+        return [Query(text=messages[-1].content)]
```
igiloh-pinecone, line 23: Please take the last message which is a `UserMessage` (`message.role == 'Role.user'` or `isinstance()`).  
If you go backwards on the messages and can't find a `UserMessage` - then raise an error

igiloh-pinecone (changes requested):
@izellevy see one comment

src/canopy/chat_engine/query_generator/last_message.py
acatav: This sounds a bit weird to me. It's more likely that if the last message is not user message, there is something wrong in the pipeline or usage of this class, so IMO it's better to raise an error here instead of silently do something that might not be expected by the user

izellevy: I see your point @igiloh-pinecone but agree with @acatav. Maybe I can raise if the history is empty or the last message is not a user message @igiloh-pinecone wdyt?

acatav (approved):
LGTM

igiloh-pinecone (approved):
LGTM

Issue #183, opened 2023-11-20
[Bug] Error on creating new index with canopy new


coolxeo:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

## Command
When I launch the command `canopy new` to create a new index I receive the following error:

```
Error: Failed to create a new index. Reason:
Failed to infer vectors' dimension from encoder due to error: 'ascii' codec can't encode character '\u201c' in position 7: ordinal not in range(128). Please fix the error or provide the dimension manually
```

### Expected Behavior

A new index is created

## Context
- In a fresh installation of canopy
- With a Pinecone account with no indexes
- In Mac Os Sonoma 14.1
- With Python 3.11.4

### Relevant log output

```shell
canopy new
Canopy is going to create a new index: canopy--merchants
Do you want to continue? [y/N]: y
Error: Failed to create a new index. Reason:
Failed to infer vectors' dimension from encoder due to error: 'ascii' codec can't encode character '\u201c' in position 7: ordinal not in range(128). Please fix the error or provide the dimension manually
```
```


### Environment

```markdown
- **OS**: Mac Os Sonoma 14.1
- **Language version**: Python 3.11.4
- **Pinecone client version**: Canopy, version 0.2.0
```

igiloh-pinecone:
Thank you very much @coolxeo for reporting this problem, and apologies for the inconvenience! 

We are  trying to reproduce the problem so we can debug it.  
I'm suspecting this might be locale related, can you please run
```python
import locale
print(locale.getdefaultlocale())
print(locale.getlocale())
```
On the same environment and post the result here?

coolxeo:
@igiloh-pinecone I just found out what it was!
when copying pasting the env variables it copy to my console a strange quote character
`export PINECONE_ENVIRONMENT=“gcp-starter”`
I replaced with the proper one
`export PINECONE_ENVIRONMENT="gcp-starter"`

I saw this before with Mac, apologise for the inconvenient!

PR #220, opened 2023-12-12
Slightly improve error handling for external providers

igiloh-pinecone:
## Solution

This is a very initial and partial solution for #183.  
We will need to improve error handling much further.

## Type of Change


src/canopy/knowledge_base/knowledge_base.py
```diff
@@ -307,16 +307,19 @@ def create_canopy_index(self,
...
+                    )
+                dimension = encoder_dimension
             except Exception as e:
                 raise RuntimeError(
-                    f"Failed to infer vectors' dimension from encoder due to error: "
-                    f"{e}. Please fix the error or provide the dimension manually"
+                    f"Canopy has failed to infer vectors' dimensionality using the "
+                    f"selected encoder: {self._encoder.__class__.__name__}. You can "
+                    f"provide the dimension manually, try using a different encoder, or"
+                    f" fix the underlying error:\n{e}"
```
miararoy, line 321: this part should have system test

miararoy (dismissed):
add system test, minimal, overall i think this is pretty thorough as-is, but I understand the notion of this can be extended a lot

src/canopy/knowledge_base/knowledge_base.py
igiloh-pinecone: It has. And you're right - I need to fix that test

src/canopy/knowledge_base/record_encoder/openai.py
```diff
@@ -27,25 +36,31 @@ def __init__(self,
...
-
     async def _aencode_documents_batch(self,
                                        documents: List[KBDocChunk]
                                        ) -> List[KBEncodedDocChunk]:
         raise NotImplementedError
 
     async def _aencode_queries_batch(self, queries: List[Query]) -> List[KBQuery]:
         raise NotImplementedError
+
+    def _format_error(self, err):
```
acatav, line 57: Maybe we should resuse this code in the LLM too? I think it's can improve the experience

src/canopy/knowledge_base/record_encoder/openai.py
```diff
@@ -27,25 +36,31 @@ def __init__(self,
                         Defaults to 400.
             **kwargs: Additional arguments to pass to the underlying `pinecone-text. OpenAIEncoder`.
         """  # noqa: E501
-        encoder = OpenAIEncoder(model_name, **kwargs)
+        try:
+            encoder = OpenAIEncoder(model_name, **kwargs)
+        except OpenAIError as e:
+            raise RuntimeError(
```
acatav, line 42: Probably for the future, but I guess it's reasonable to have our own auth and rate limit errors. For many services we are going to use this issue might occur, even pinecone by itself

src/canopy/llm/openai.py
```diff
@@ -175,15 +197,22 @@ def enforced_function_call(self,
...
+                tools=[function_dict],
+                tool_choice={"type": "function",
+                             "function": {"name": function.name}},
+                max_tokens=max_tokens,
+                **model_params_dict
+            )
+        except openai.OpenAIError as e:
+            provider_name = self.__class__.__name__.replace("LLM", "")
+            raise RuntimeError(
+                f"Failed to use {provider_name}'s {self.model_name} model for chat completion.\n"
```
acatav, line 213: I think it's better to have this message indicate that the error is coming from the function calling method and not for the chat completion method, it will be easier for debugging

src/canopy_cli/cli.py
```diff
@@ -140,9 +134,12 @@ def _validate_chat_engine(config_file: Optional[str]):
...
     try:
-        ChatEngine.from_config(config.get("chat_engine", {}))
-    except openai.OpenAIError:
-        raise CLIError(OPENAI_AUTH_ERROR_MSG)
+        # If the server itself will fail, we can't except the error, since it's running
+        # in a different process. Try to load and run the ChatEngine so we can catch
+        # any errors and print a nice message.
+        chat_engine = ChatEngine.from_config(config.get("chat_engine", {}))
+        chat_engine.context_engine.knowledge_base.connect()
+        chat_engine.chat([UserMessage(content="hello")])
```
acatav, line 142: better to change to a low `max_tokens` so the latency will be much lower

acatav (approved):
LGTM

src/canopy/knowledge_base/record_encoder/openai.py
igiloh-pinecone: Yeah, that's the plan. We'll do it as part of the planned `BaseLLM` refactor anyway

src/canopy/knowledge_base/record_encoder/openai.py
igiloh-pinecone: Yeah, that's definitely the plan!

PR #226, opened 2023-12-14
Add instruction query generator

izellevy:
## Problem

Currently, we only support LLMs that have function calling functionality for query generation. We need to support LLMs that do not have this functionality.

## Solution

Added a new query generator that is able to generate a question based on the chat history without using function calling

## Type of Change

## Test Plan

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1205411199043212


config/anyscale.yaml
```diff
@@ -27,7 +27,7 @@ chat_engine:
   # Since Anyscale's LLM endpoint currently doesn't support function calling, we will use the LastMessageQueryGenerator
   # --------------------------------------------------------------------
   query_builder:
-    type: LastMessageQueryGenerator     # Options: [FunctionCallingQueryGenerator, LastMessageQueryGenerator]
+    type: CondensedQueryGenerator     # Options: [FunctionCallingQueryGenerator, LastMessageQueryGenerator]
```
igiloh-pinecone, line 30: We should list `CondensedQueryGenerator` as an option in the default `config.yaml` as well

src/canopy/chat_engine/query_generator/condensed.py
```diff
@@ -0,0 +1,139 @@
...
+    def __init__(self,
+                 *,
+                 llm: Optional[BaseLLM] = None):
+        """
+             Tries to build a question out of all the chat history.
+             In order to do that, the system first tries to make LLM return
+             a JSON object with a key "question" and the value representing
+             the question the user was intending to ask. If LLM response
+             cannot be parsed it falls back to last message query generator,
+             meaning it returns the last message of the history as a query.
```
igiloh-pinecone, line 74:
```suggestion
             This `QueryGenerator` uses an LLM to formulate a knowledge base query from the full chat history.
             It does so by prompting the LLM to reply with a JSON containing a single key `question`, containing the
             If LLM the response cannot be parsed (after multiple retries), it falls back to returning the last message
             from the history as a query, much like `LastMessageQuerGenerator`
```

src/canopy/chat_engine/query_generator/condensed.py
```diff
@@ -0,0 +1,139 @@
...
+from canopy.chat_engine.query_generator import QueryGenerator, LastMessageQueryGenerator
+from canopy.llm import BaseLLM, AnyscaleLLM
+from canopy.models.data_models import Messages, Query, UserMessage
+from canopy.models.api_models import ChatResponse
+
+logger = logging.getLogger(__name__)
+
+SYSTEM_PROMPT = """
+You are an expert on formulating a search query for a search engine, to assist
+ in responding to the user's question.
```
igiloh-pinecone, line 18:
```suggestion
SYSTEM_PROMPT = """You are an expert on formulating a search query for a search engine, to assist
 in responding to the user's question.
```

You shouldn't drop down a line in the beginning if a `"""` block that is sent to the LLM.  
From the LLM's perspective - tour system prompt now start with a `\n` token, and it has been shown that for some LLMs that completely changes the end result... 🤷‍♂️

src/canopy/chat_engine/query_generator/condensed.py
```diff
@@ -0,0 +1,139 @@
...
+             In order to do that, the system first tries to make LLM return
+             a JSON object with a key "question" and the value representing
+             the question the user was intending to ask. If LLM response
+             cannot be parsed it falls back to last message query generator,
+             meaning it returns the last message of the history as a query.
+        """
+        self._llm = llm or self._DEFAULT_COMPONENTS["llm"]()
+        self._system_prompt = SYSTEM_PROMPT
+        self._prompt_builder = PromptBuilder(HistoryPruningMethod.RAISE, 2)
+        self._last_message_query_generator = LastMessageQueryGenerator()
```
igiloh-pinecone, line 79: Why compilation in this case and not inheritance?  

In this particular case, I think that inheriting `LastMessageQueryGenerator()` class, then in the fallback case calling `super().generate()` is the correct approach, isn't it?

src/canopy/chat_engine/query_generator/condensed.py
```diff
@@ -0,0 +1,139 @@
...
+        # Define a regex pattern to find the JSON object with the key "question"
+        self._question_regex = re.compile(r'{\s*"question":\s*"([^"]+)"\s*}')
+
+    def generate(self,
+                 messages: Messages,
+                 max_prompt_tokens: int) -> List[Query]:
+
+        # Add a user message at the end; that helps us return a JSON object.
+        new_history = (
+                messages +
```
igiloh-pinecone, line 90: We should probably make this configurable - e.g. by adding a `append_instruction_reminder_to_history: bool = True` parameter. (Feel free to find a shorter name... 😄)

It might be that some LLMs don't respond well to this additional message, and it only confuses them.

tests/unit/query_generators/test_condensed_query_generator.py
```diff
@@ -0,0 +1,85 @@
...
+                AssistantMessage(content="Which kind of client?"),
+                UserMessage(content="A pinecone client.")]
+
+    @staticmethod
+    @pytest.mark.parametrize(("response", "query"), [
+        (
+                '{"question": "How do I init a pinecone client?"}',
+                "How do I init a pinecone client?"),
+        (
+                '...',
```
igiloh-pinecone, line 40: I'm not sure I'm getting this expected result (`'...'`), can you please explain?

tests/unit/query_generators/test_condensed_query_generator.py
```diff
@@ -0,0 +1,85 @@
...
+                )
+            ]
+        )
+
+        result = query_generator.generate(messages=sample_messages,
+                                          max_prompt_tokens=4096)
+
+        assert isinstance(result, List)
+        assert len(result) == 1
+        assert result[0] == Query(text=query)
```
igiloh-pinecone, line 75: You should probably verify that in the non-paraeable case, `mock_llm.chat_completion()` was called 3 times.

igiloh-pinecone (commented):
@izellevy please see a few suggestions

src/canopy/chat_engine/query_generator/condensed.py
izellevy: @igiloh-pinecone the inputs to the init function are completely different and I don't see any connection between them. It does not extend the functionality but more like calling the other in case it cannot generate. So I would want to keep it this way. Lmk what you think

tests/unit/query_generators/test_condensed_query_generator.py
izellevy: I wanted to simulate that if the model return a non-parsable json we receive the last message as an answer. I will change it to something more understandable.

acatav (commented):
LGTM overall! left some comments for your decision, and some for a discussion

src/canopy/chat_engine/query_generator/condensed.py
izellevy: OpenAI and Anyscale LLMs respond it quite well, I think in any case there is a lot of logic tied to this query generator so not sure customizing only that part will give us more flexibility. We will probably put a parameter that no one will use. In the future maybe we can improve this to template the user prompt & system prompt and give complete flexibility to the user but for now I wanna keep it as it is. LMK what you think

src/canopy/chat_engine/query_generator/condensed.py
```diff
@@ -0,0 +1,139 @@
...
+"""  # noqa: E501
+
+USER_PROMPT = "Return only a JSON that has 'question' as a key and the value."
+
+
+class ExtractionException(ValueError):
+    pass
+
+
+class CondensedQueryGenerator(QueryGenerator):
```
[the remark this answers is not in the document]
izellevy: @igiloh-pinecone what do you think about the name?

src/canopy/chat_engine/query_generator/condensed.py
```diff
@@ -0,0 +1,139 @@
...
+    _DEFAULT_COMPONENTS = {
+        "llm": AnyscaleLLM,
```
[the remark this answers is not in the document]
izellevy: I also think so but @igiloh-pinecone had a comment on this. I agree that our default should be OpenAI

src/canopy/chat_engine/query_generator/condensed.py
igiloh-pinecone: We'll agree to disagree 😄 .

Sure, no problem. Keep it this way.

src/canopy/chat_engine/query_generator/condensed.py
igiloh-pinecone: Ok, agreed.

src/canopy/chat_engine/query_generator/condensed.py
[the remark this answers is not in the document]
igiloh-pinecone: I tend to agree with @acatav , the name `Instruction` or even `Prompt` is more common.  
The "condensed" naming is a term that LLamaIndex invented, and I haven't seen it used otherwise.

igiloh-pinecone (approved):
LGTM!  

I do agree with @acatav about the name though. @izellevy - up to you

PR #227, opened 2023-12-14
Adding function calling to AnyscaleLLM

kylehh:
## Problem

1. Enable function calling to Anyscale LLM
2. Add default {'type'='array'} for `FunctionArrayProperty`
## Solution

Describe the approach you took. Link to any relevant bugs, issues, docs, or other resources.

## Type of Change

## Test Plan


src/canopy/llm/anyscale.py
```diff
@@ -41,23 +33,6 @@ def __init__(
...
         super().__init__(model_name, api_key=ae_api_key, base_url=ae_base_url, **kwargs)
 
-    @retry(
-        reraise=True,
-        stop=stop_after_attempt(3),
-        retry=retry_if_exception_type(
-            (json.decoder.JSONDecodeError, jsonschema.ValidationError)
-        ),
-    )
-    def enforced_function_call(
```
igiloh-pinecone, line 51: I think it would be wise to wrap the parent's class function with a  `try: except:`.  
Then in case where user configured the wrong model (e.g. llama2) - we would catch the error coming from Anyscale, and re-raise with a clear actionable error message like `f"Model {self.model_name} doesn't support function calling. To use function calling capability, please select a different `model_name` in configuration.`

src/canopy/llm/anyscale.py
igiloh-pinecone, line 1: @kylehh Can you please add:
1. Relevant test cases in `test_anyscale.py`
2. Change the `anyscale.yaml` config file to use `FunctionCallingQueryGenerator` with mistral

igiloh-pinecone (changes requested):
@kylehh All in all LGTM. Please see a few comments

src/canopy/llm/anyscale.py
kylehh: done

src/canopy/llm/models.py
```diff
@@ -20,6 +20,7 @@ class FunctionArrayProperty(BaseModel):
     def dict(self, *args, **kwargs):
         super_dict = super().dict(*args, **kwargs)
         if "items_type" in super_dict:
+            super_dict["type"] = "array"
```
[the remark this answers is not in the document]
kylehh: Yes, I run pytest on OAI and it works with this change
FYI, the JSON schema reference [here](https://json-schema.org/understanding-json-schema/reference/array#array)

tests/system/llm/test_anyscale.py
```diff
@@ -31,29 +36,44 @@ class TestAnyscaleLLM:
     @staticmethod
     @pytest.fixture
     def model_name():
-        return "meta-llama/Llama-2-7b-chat-hf"
```
kylehh, line 34: Change default model to "mistral" which supports both chat and function calling

tests/system/llm/test_anyscale.py
```diff
@@ -31,29 +36,44 @@ class TestAnyscaleLLM:
...
+        return "mistralai/Mistral-7B-Instruct-v0.1"
```
kylehh, line 39: Change default model to "mistral" which supports both chat and function calling

tests/system/llm/test_anyscale.py
```diff
@@ -31,29 +36,44 @@ class TestAnyscaleLLM:
...
     @staticmethod
     @pytest.fixture
     def messages():
         # Create a list of MessageBase objects
         return [
-            MessageBase(role=Role.USER, content="Hello, assistant."),
-            MessageBase(
-                role=Role.ASSISTANT, content="Hello, user. How can I assist you?"
-            ),
+            MessageBase(role=Role.SYSTEM, content="You are a helpful AI assistant."),
```
kylehh, line 46: more standard message format, SYSTEM msg followed by a USER msg. ( messages ending with ASSISTANT msg would error out with AE function calling)

src/canopy/llm/anyscale.py
kylehh: Good suggestion. Added model check for function calls

tests/system/llm/test_anyscale.py
igiloh-pinecone, line 1: @kylehh we've changed the format of `test_openai.py` in a manner where it parameterizes all LLM classes that inherit from `OpenAILLM` (like `AnyscaleLLM`), and runs the same test cases for all of these classes.  
This way - we don't need a dedicated `test_anyscale.py` file which simply duplicates the entire test cases

igiloh-pinecone (approved):
LGTM!

I've changed the testing scheme and made `AnyscalseLLM` another parameterized case in `test_openai.py`

PR #232, opened 2023-12-19
Cleanup indexes in case of failure

izellevy:
## Problem

When an action get cancelled pytest is stopped using SIGKILL (by Github CI) which causes unused indexes to remain after the test.

## Solution

Add a new step to clean up the indexes in case the run is interrupted in the middle.

## Type of Change

## Test Plan

### DIFF
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -96,7 +96,7 @@ jobs:
         echo "Test Run ID: ${run_id}"
         poetry run pytest -n 3 --dist loadscope --testrunuid $run_id --html=report_e2e.html --self-contained-html tests/e2e
     - name: Cleanup indexes
-      if: cancelled() || failure() #&& github.event_name == 'merge_group'
+      if: cancelled() || failure() # && github.event_name == 'merge_group'
       env:
         PINECONE_ENVIRONMENT: ${{ matrix.pinecone-plan == 'paid' && secrets.PINECONE_ENVIRONMENT_3 || secrets.PINECONE_ENVIRONMENT_4 }}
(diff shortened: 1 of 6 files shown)


.github/workflows/PR.yml
```diff
@@ -84,7 +90,25 @@ jobs:
...
         CO_API_KEY: ${{ secrets.CO_API_KEY }}
         CE_LOG_FILENAME: e2e.log
-      run: poetry run pytest -n 3 --dist loadscope --html=report_e2e.html --self-contained-html tests/e2e
+      run: |
+        run_id=$(uuidgen | tr -d '-' | tr '[:upper:]' '[:lower:]')
+        poetry run pytest -n 3 --dist loadscope --testrunuid $run_id --html=report_e2e.html --self-contained-html tests/e2e
+        echo "Test Run ID: ${run_id}"
+        echo "run_id=${run_id}" >> $GITHUB_OUTPUT
+    - name: Cleanup indexes
+      if: failure() && github.event_name == 'merge_group'
```
igiloh-pinecone, line 99:
```suggestion
      if: always() && github.event_name == 'merge_group'
```
Let's not take any chances. The added time for successful tests is probably negligible.

igiloh-pinecone (approved):
@izellevy see one comment, other than that LGTM

scripts/cleanup_indexes.py
```diff
@@ -0,0 +1,23 @@
...
+
+def main():
+    logging.basicConfig(level=logging.INFO)
+    logger = logging.getLogger(__name__)
+
+    if len(sys.argv) != 2:
+        logger.info("Usage: python scripts/cleanup_indexes.py <testrun_uid>")
+        sys.exit(1)
+
+    testrun_uid = sys.argv[1]
```
igiloh-pinecone, line 14: Please promise me that if we're every adding even more one parameter, we're switching to `argparse` 😄

igiloh-pinecone (approved):
LGTM!

PR #234, opened 2023-12-21
Add dockerfile

izellevy:
## Problem

 Build/deploy of canopy is not streamlined.

## Solution

Add a dockerfile to host canopy.

## Explanation

This Dockerfile defines a multi-stage build process for a Python application using Poetry for dependency management. Let's break down each stage and understand what it does:

### Stage 1: Python Base
This stage sets up the shared environment variables and creates a virtual environment using the official Python 3.11.7-slim image. Key environment variables include settings for Python, pip, and Poetry. The PYTHONPATH is also configured. The virtual environment is created at /venv.

### Stage 2: Builder Base
This stage extends the Python Base stage and adds tools necessary for building dependencies. It installs additional packages using apt-get such as build essentials, Git, Vim, and other dependencies. Poetry is installed, and the build cache is utilized to speed up the build process. It sets the working directory to /app and copies pyproject.toml and poetry.lock to install dependencies.

### Stage 3: Development
This stage is used during development/testing. It extends the Builder Base stage and copies the project files. It installs development dependencies, and the CMD is set to run a Bash shell.

### Stage 4: Production
This is the final stage used for runtime. It extends the Python Base stage and copies the built Poetry and virtual environment from the Builder Base stage. It installs runtime dependencies, copies the project files, and sets up the environment for production. The application is exposed on port 8000, and Gunicorn is configured as the entry point to run the application.

**_**Optimizations and Considerations:**_**
### Multi-stage Build:

The use of multi-stage builds helps to keep the final image small by discarding unnecessary build dependencies.
The production image only includes the necessary artifacts for runtime, reducing its size.
### Build Cache:

Caching is used effectively during the installation of dependencies to speed up the build process.
Poetry and pip caches are stored in the build cache directory to allow reuse.
### Virtual Environment:

A virtual environment is used for dependency isolation.
The virtual environment is created in a separate stage to ensure a clean environment and is then copied to the production image.

### Dependency Caching Optimization
As part of the Dockerfile, the optimization of installing dependencies with the --no-root option is employed. This is mentioned in both the Builder Base and Development stages. The --no-root option allows for the caching of dependencies, as they are installed in a location that is not the final root of the system. This can significantly speed up subsequent builds, as the dependencies are cached separately from the application code.

Here's the specific part in the Dockerfile where this optimization is implemented:

#### Builder Base Stage
RUN --mount=type=cache,target=/root/.cache \
    poetry install --no-root --all-extras --only main

#### Development Stage
RUN --mount=type=cache,target=/root/.cache \
    poetry install --no-root --all-extras --with dev

## Type of Change

## Test Plan

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,7 +58,8 @@ RUN --mount=type=cache,target=/root/.cache \
 
 # used to init dependencies
 WORKDIR /app
-COPY poetry.lock pyproject.toml ./
+COPY pyproject.toml ./
+RUN poetry lock
 
 # install runtime deps to VIRTUAL_ENV
 RUN --mount=type=cache,target=/root/.cache \
@@ -72,7 +73,9 @@ RUN --mount=type=cache,target=/root/.cache \
 FROM builder-base as development
 
 WORKDIR /app
-COPY poetry.lock pyproject.toml ./
+COPY --from=builder-base /app/pyproject.toml pyproject.toml
+COPY --from=builder-base /app/poetry.lock poetry.lock
+
 
 # quicker install as runtime deps are already installed
 RUN --mount=type=cache,target=/root/.cache \
@@ -105,7 +108,8 @@ COPY --from=builder-base $VIRTUAL_ENV $VIRTUAL_ENV
 
 WORKDIR /app
 
-COPY poetry.lock pyproject.toml ./
+COPY --from=builder-base /app/pyproject.toml pyproject.toml
+COPY --from=builder-base /app/poetry.lock poetry.lock
 
 COPY src/ src/
 COPY config/ config/
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,13 +59,11 @@ RUN --mount=type=cache,target=/root/.cache \
 # used to init dependencies
 WORKDIR /app
 COPY poetry.lock pyproject.toml ./
-COPY src/ src/
-COPY config/ config/
-
 
 # install runtime deps to VIRTUAL_ENV
 RUN --mount=type=cache,target=/root/.cache \
     poetry install --no-root --all-extras --only main
+
 
 ################################
 # DEVELOPMENT
@@ -74,10 +72,15 @@ RUN --mount=type=cache,target=/root/.cache \
 FROM builder-base as development
 
 WORKDIR /app
+COPY poetry.lock pyproject.toml ./
 
 # quicker install as runtime deps are already installed
 RUN --mount=type=cache,target=/root/.cache \
     poetry install --no-root --all-extras --with dev
+
+COPY . .
+RUN poetry install --all-extras --only-root
+
 
 EXPOSE 8000
 CMD ["bash"]
@@ -88,8 +91,8 @@ CMD ["bash"]
 # Final image used for runtime
 ################################
 FROM python-base as production
-ENV WORKER_COUNT=1
 
+ENV WORKER_COUNT=1
 
 RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
     apt-get install -y --no-install-recommends \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,119 @@
...
+    ca-certificates && \
+    apt-get clean
+
+# copy in our built poetry + venv
+COPY --from=builder-base $POETRY_HOME $POETRY_HOME
+COPY --from=builder-base $VIRTUAL_ENV $VIRTUAL_ENV
+
+WORKDIR /app
+
+COPY poetry.lock pyproject.toml ./
```
igiloh-pinecone, line 108: Again - you will need to copy `poetry.lock` from builder-base, since its not pushed to the repo

Dockerfile
```diff
@@ -0,0 +1,119 @@
...
+# The --mount will mount the buildx cache directory to where
+# Poetry and Pip store their cache so that they can re-use it
+RUN --mount=type=cache,target=/root/.cache \
+    curl -sSL https://install.python-poetry.org | python -
+
+# used to init dependencies
+WORKDIR /app
+COPY poetry.lock pyproject.toml ./
+
+# install runtime deps to VIRTUAL_ENV
```
igiloh-pinecone, line 63: Maybe we should cache the entire `venv/`?  
I mean - between two builds with the same `pyproject.toml`, there should be literally zero installations, right?

Dockerfile
```diff
@@ -0,0 +1,119 @@
...
+RUN --mount=type=cache,target=/root/.cache \
+    poetry install --no-root --all-extras --only main
```
igiloh-pinecone, line 65: Why `--all-extras` and `--only main` together?

Dockerfile
```diff
@@ -0,0 +1,119 @@
...
+COPY --from=builder-base $POETRY_HOME $POETRY_HOME
+COPY --from=builder-base $VIRTUAL_ENV $VIRTUAL_ENV
+
+WORKDIR /app
+
+COPY poetry.lock pyproject.toml ./
+
+COPY src/ src/
+COPY config/ config/
+RUN touch README.md
```
igiloh-pinecone, line 112: Why do we need this?

Makefile
```diff
@@ -0,0 +1,61 @@
...
+.PHONY: lint static test test-unit test-system test-e2e docker-build docker-build-dev docker-run docker-run-dev help
+
+lint:
+	poetry run flake8 .
+
+static:
+	poetry run mypy src
+
+test:
+	poetry run pytest -n 8 --dist loadscope
```
igiloh-pinecone, line 18: The number of workers should probably be a param (with default)

Makefile
```diff
@@ -0,0 +1,61 @@
+IMAGE_NAME = canopy
+DOCKERFILE_DIR = .
+COMMON_BUILD_ARGS = --progress plain
+EXTRA_BUILD_ARGS =
```
igiloh-pinecone, line 4: Can you please make it possible to pass these as arguments directly to `make build`, not only as env vars?

igiloh-pinecone (changes requested):
Please see a few suggestions

Makefile
izellevy: Done

Makefile
```diff
@@ -5,6 +5,7 @@ EXTRA_BUILD_ARGS =
 IMAGE_VERSION = $(shell poetry version -s)
 PORT = 8000
 ENV_FILE = .env
+TEST_WORKER_COUNT = 8
```
igiloh-pinecone, line 8: Are spaces after `=` allowed in Makefiles?

igiloh-pinecone (approved):
LGTM!

Makefile
izellevy: Yep 🙂

Issue #225, opened 2023-12-14
[Bug] Canopy Chat Not Working


DanKamNdi:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

The server starts in the terminal. But when I do `canopy chat` it doesn't do anything. It is just blank. It doesn't crash or do anything.

![image](https://github.com/pinecone-io/canopy/assets/16070606/f1e7788e-8bca-436d-ab1c-379f95de7dfe)

The terminal just stays blank.
![image](https://github.com/pinecone-io/canopy/assets/16070606/82420a9c-d4d0-4045-9a2f-819a8de4fc40)

### Expected Behavior

This is what should happen

![image](https://github.com/pinecone-io/canopy/assets/16070606/971b6423-91b7-4407-aaa7-b0a8a548ff0b)

### Steps To Reproduce

I followed the setup. [](url)

### Environment

```markdown
- **OS**: Windows 11
- **Language version**: Python 3.11.4
- **Canopy version**: 0.5.0
```

miararoy:
@hi @DanKamNdi 

Looking into that and will respond soon, thanks for flagging this!

miararoy:
hi, @DanKamNdi 

Just letting you know we are still working on this

izellevy:
@DanKamNdi looks like uvicorn in Windows closes the receiving socket after a certain amount of time. Since we do not have a timeout for the health check (in canopy chat route), the request hangs. In any case, I will open a PR to raise a proper error in case we reach a read timeout from the canopy server but is it possible for you to try restarting the server and let me know if it works?

izellevy:
See https://github.com/pinecone-io/canopy/pull/236

izellevy:
Also will keep the issue open until we understand/solve why the server stops responding after a specific time. @igiloh-pinecone @miararoy FYI

PR #236, opened 2023-12-25
Add timeout to checking server health

izellevy:
## Problem

[See](https://github.com/pinecone-io/canopy/issues/225) 

## Solution

Added a timeout for health checks

## Type of Change

## Test Plan


src/canopy_cli/cli.py
```diff
@@ -200,18 +207,18 @@ def health(url):
...
-
-        A Canopy service cannot be started without a Pinecone index that is configured
-        to work with Canopy. This command creates a new Pinecone index and configures
-        it in the right schema.
-
-        If the embedding vectors' dimension is not explicitly configured in
-        the config file, the embedding model will be tapped with a single token to
-        infer the dimensionality of the embedding space.
-        """  # noqa: E501
+            """
```
igiloh-pinecone, line 210: Please don't push these code style changes as part of this commit. It can only lead to conflicts with other open PRs.  
Code style changes (to lines unaffected by the commit's purpose) can be done via a separate dedicated PR

src/canopy_cli/cli.py
```diff
@@ -67,6 +65,14 @@ def check_server_health(url: str):
         )
         raise CLIError(msg)
 
+    except requests.exceptions.ReadTimeout:
+        msg = f"""
+        Canopy server did not send any data in the allotted
+        amount of time ({timeout_seconds} seconds).
+        Try restarting the server.
```
igiloh-pinecone, line 72: Don't write "try restarting the server", that's not a viable solution.  
Please change it to something like "Please check that the server is running and that the provided url {url} is correct", or something of that sort.

The problem with the server getting "stuck" on Windows is a separate issue, which we should fix in its root cause.

igiloh-pinecone (commented):
@izellevy all in all LGTM. Please see a few minor comments.

Also - either change the PR description so it won't close the Windows bug, please open a new issue to track the real problem on windows (server stops responding)

src/canopy_cli/cli.py
izellevy: Done

izellevy:
@igiloh-pinecone removed the PR dependency. Keeping the original issue for tracking.

igiloh-pinecone (commented):
LGTM

PR #242, opened 2024-01-03
Build and Publish Docker Image

izellevy:
## Purpose

The purpose is to publish docker images on each merge to main and each release.

## Solution

Added a new workflow file that builds and pushes the image to GitHub packages.

## Type of Change

## Test Plan

### DIFF
--- a/.github/workflows/build-push-image.yml
+++ b/.github/workflows/build-push-image.yml
@@ -1,0 +1,40 @@
+name: Create and publish a Docker image
+
+on:
+  workflow_dispatch:
+    inputs:
+      image-tag:
+        type: string
+        required: true
+        description: 'Docker image tag'
+env:
+  REGISTRY: ghcr.io
+  REPOSITORY: ${{ github.repository }}
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Log in to the Container registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push
+        uses: docker/build-push-action@v5
+        env:
+          IMAGE_TAG: ${{ inputs.image-tag || `git describe --tags | sed 's/V//'` }}
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-push-image.yml
```diff
@@ -0,0 +1,48 @@
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Docker metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+              ghcr.io/pinecone-io/canopy
+          tags: |
+            type=ref,event=branch
+            type=semver,pattern={{version}}
```
igiloh-pinecone, line 37: Doesn't "version" need to be under an `if:` only for pushes to git tag?  
Otherwise we'll keep overwriting the version tag for any merge to `main`, won't we?

.github/workflows/build-push-image.yml
```diff
@@ -0,0 +1,48 @@
...
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Docker metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+              ghcr.io/pinecone-io/canopy
+          tags: |
```
igiloh-pinecone, line 35: Don't we want to add a `:latest` tag pointing to the latest stable version, and a `:nightly` tag pointing at latest push?

.github/workflows/build-push-image.yml
igiloh-pinecone, line 37: Doesn't "version" need to be under an `if:` only for pushes to git tag?  
Otherwise we'll keep overwriting the version tag for any merge to `main`, won't we?

igiloh-pinecone (commented):
@izellevy all in all LGTM.  
See two small comments

.github/workflows/build-push-image.yml
izellevy: Latest will be automatically generated (https://github.com/docker/metadata-action?tab=readme-ov-file#latest-tag) Basically it will not be called nightly but every merge to main will deploy with docker tag main (override). We can also move to nightly with cron jobs docker metadata action has support for it but I don't think our PR pace justifies that as of now.

.github/workflows/build-push-image.yml
izellevy: Basically this conditional logic is being handled on docker metadata action. If the branch is merged to main. image:main will be overridden. If there is a new tag we will create to images image:latest and image:0.x.x

igiloh-pinecone: Oh... gotcha!

igiloh-pinecone (approved):
LGTM!

@izellevy we need to think how are we going to add tests for this pipeline

PR #243, opened 2024-01-04
Add support for namespaces

### LINKED ISSUE #219: [Namespace] Using namespaces in chat engine
### Is this your first time submitting a feature request?

- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing functionality

### Describe the feature

from canopy.knowledge_base import KnowledgeBase
kb = KnowledgeBase(index_name="{{TestIndex}}",namespace="{{TestNamespace}}")
kb.connect()

Similar to chain using namespace in vector stores, namespaces should be supported in Chat Engines

### Describe alternatives you've considered

from canopy.knowledge_base import KnowledgeBase
from canopy.chat_engine import ChatEngine
kb = KnowledgeBase(index_name="demo")
kb.connect()
ContextEngine(kb,global_metadata_filter={'key': {'$eq': 'value'}})
chat_engine=ChatEngine(context_engine)

### Who will this benefit?

Balancing costs of using a new index for each set instead of multiple namespaces in an index. Metadata filtering works but is not backwards compatible with existing namespaces code. 

### Are you interested in contributing this feature?

No

### Anything else?

_No response_

izellevy:
## Problem

We don't have a chat endpoint for specific namespaces of an index.

## Solution

Added namespace support to ChatEngine, ContextEngine & KnowledgeBase
Dropped support for endpoints not containing v1 urls.

## Type of Change

## Test Plan

Added relevant tests.

Closes #219


src/canopy/chat_engine/chat_engine.py
```diff
@@ -227,9 +229,10 @@ def chat(self,
 
     def _get_context(self,
                      messages: Messages,
+                     namespace: Optional[str] = None
                      ) -> Context:
         queries = self._query_builder.generate(messages, self.max_prompt_tokens)
-        context = self.context_engine.query(queries, self.max_context_tokens)
+        context = self.context_engine.query(queries, self.max_context_tokens, namespace)
```
igiloh-pinecone, line 235:
```suggestion
        context = self.context_engine.query(queries, self.max_context_tokens, namespace=namespace)
```
We're trying to be very explicit about keyword args, and rely on argument ordering only 1-2 mandatory args.

src/canopy/knowledge_base/knowledge_base.py
```diff
@@ -401,7 +399,8 @@ def delete_index(self):
 
     def query(self,
               queries: List[Query],
-              global_metadata_filter: Optional[dict] = None
+              global_metadata_filter: Optional[dict] = None,
+              namespace: Optional[str] = None
```
igiloh-pinecone, line 403: Wait, @izellevy, this is incorrect.  
At the `KnowledgeBase` layer, each Query should still have its own individual `namespace`. The `KnowledgeBase` is used at the same "level" like the pinecone client, with the only difference that it takes data (`Document` \ `Query`) as input instead of vectors.  
Just like the same pinecone client object can handle traffic to multiple namespaces, the same `KnowledgeBase` can do the same.

Instead of deleting `namespace` from the `Query` data model, we should create a new `Query` API model (in `api_models.py`). Then, either at the `ChatEngine` layer or even the `app` layer, we should do something like:
```python
from api_models import Query 
from canopy.models.data_models import Query as InnerQuery
...
def chat(queries: List[Query], ..., namespace=None,...):
    queries = [InnerQuery(**q, namespace=namespace) for q in queries]
     ....
```

The code of `KnowledgeBase` shouldn't change at all.  
Whether the `ContextEngine` should allow namespace per query is debatable. I tend to say it does.

@miararoy WDYT?

igiloh-pinecone (changes requested):
@izellevy Either I recall our final decision incorrectly, or this is not the intended design.

At the `KnowledgeBase` level, namespaces should still be passed per query.  
Only for the outer most API layer (the `app.py`, and maybe the `ChatEngine` object) we want to enforce a single namespace route.

src/canopy/knowledge_base/knowledge_base.py
```diff
@@ -417,6 +416,8 @@ def query(self,
             global_metadata_filter: A metadata filter to apply to all queries, in addition to any query-specific filters.
                                     For example, the filter {"website": "wiki"} will only return documents with the metadata {"website": "wiki"} (in case provided in upsert)
                                     see https://docs.pinecone.io/docs/metadata-filtering
+            namespace: The namespace of the index for context retreival. To learn more about namespaces, see https://docs.pinecone.io/docs/namespaces
```
igiloh-pinecone, line 419:
```suggestion
            namespace: The namespace that will be queried in the underlying index. To learn more about namespaces, see https://docs.pinecone.io/docs/namespaces
```

src/canopy_server/app.py
```diff
@@ -95,13 +95,19 @@
 logger: logging.Logger
```
igiloh-pinecone, line 97: @izellevy you also need to add `/{namespace}/` param to the `/upsert` and `/query` routes.  
As we discussed - for those two routes it can be at the end, no need to add another dedicated route (e.g. `/query/{namespace}`)

igiloh-pinecone (changes requested):
A few more comments

src/canopy_server/app.py
izellevy: I think it is better to keep the same convention from the users perspective. So I prepended the namespace for now. Let's take it offline

src/canopy/context_engine/context_engine.py
```diff
@@ -81,13 +81,16 @@ def __init__(self,
 
         self.global_metadata_filter = global_metadata_filter
 
-    def query(self, queries: List[Query], max_context_tokens: int, ) -> Context:
+    def query(self, queries: List[Query],
+              max_context_tokens: int,
+              namespace: Optional[str] = None) -> Context:
```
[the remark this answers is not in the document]
izellevy: Done

src/canopy_server/app.py
```diff
@@ -288,13 +313,8 @@ async def startup():
 
 
 def _init_routes(app):
-    # Include the application level router (health, shutdown, ...)
-    app.include_router(application_router, include_in_schema=False)
```
igiloh-pinecone, line 292: We need to remember to say in the next release's changelog that routes without `/v1` aren't supported anymore.

tests/unit/chat_engine/test_chat_engine.py
igiloh-pinecone, line 1: Please add a few more test cases also in `test_app.py` (at least for chat and query)

igiloh-pinecone (commented):
Please see a few more comments

src/canopy_cli/cli.py
```diff
@@ -413,9 +419,14 @@ def _chat(
             "Please set the OPENAI_API_KEY environment "
             "variable."
         )
+
+    if api_base is not None and namespace is not None:
+        api_base = os.path.join(api_base, namespace)
```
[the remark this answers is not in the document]
izellevy: Right, changed it 👍

src/canopy_cli/cli.py
```diff
@@ -488,7 +499,11 @@ def _chat(
 @click.option("--chat-server-url", default=DEFAULT_SERVER_URL,
               help=("URL of the Canopy server to use."
                     f" Defaults to {DEFAULT_SERVER_URL}"))
-def chat(chat_server_url, rag, debug, stream):
+@click.option("--namespace", "-n", default=None, envvar="INDEX_NAMESPACE",
```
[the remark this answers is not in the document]
izellevy: You are right, maybe it is a bit weird to enforce a namespace let's keep it dynamic. I am updating the docs also

src/canopy_server/app.py
izellevy: I will write in PR description not to forget.

src/canopy_server/app.py
```diff
@@ -155,6 +162,7 @@ def stringify_content(response: StreamingChatResponse):
 )
 async def query(
     request: ContextQueryRequest = Body(...),
+    namespace: Optional[str] = None,
 ) -> ContextResponse:
     """
     Query the knowledge base for relevant context.
```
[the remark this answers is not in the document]
izellevy: Done

tests/unit/chat_engine/test_chat_engine.py
izellevy: Done

src/canopy_server/app.py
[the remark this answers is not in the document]
izellevy: Done

tests/e2e/test_app.py
```diff
@@ -28,25 +29,37 @@
...
+    return request.param
+
+
+@pytest.fixture(scope="module")
+def namespace_prefix(namespace):
+    return f"{namespace}/" if namespace is not None else ""
+
+
+@pytest.fixture(scope="module")
+def index_name(testrun_uid: str, namespace):
```
igiloh-pinecone, line 43: Actually - creating the index _shouldn't_ depend on the `namespace` fixture.  
We should create a single index, then repeat all tests against the same index - once with namespace and once without namespace.

tests/util.py
```diff
@@ -15,15 +18,19 @@ def create_system_tests_index_name(testrun_uid: str) -> str:
     return create_index_name(testrun_uid, "test-kb")
 
 
-def create_e2e_tests_index_name(testrun_uid: str) -> str:
-    return create_index_name(testrun_uid, "test-app")
+def create_e2e_tests_index_name(testrun_uid: str,
```
igiloh-pinecone, line 21: See comment above.  
The whole point of namespaces, is that a single index can support multiple namespaces (including the "default" namespace).

igiloh-pinecone (approved):
@izellevy approved, but please see one comment about the test

tests/e2e/test_app.py
izellevy: Right fixed that!

tests/util.py
izellevy: Yeah right fixed this also.

PR #246, opened 2024-01-14
New Pinecone Client

izellevy:
## Problem

Main branch is not supporting the new Pinecone client.

## Solution

Merging 0.4.0 release branch back into main.

## Type of Change

## Test Plan


.github/workflows/PR.yml
```diff
@@ -83,11 +80,9 @@ jobs:
     - name: Run e2e tests
       id: e2e_tests
       if: github.event_name == 'merge_group'
-      continue-on-error: ${{ matrix.pinecone-plan == 'starter' }}
       env:
         INDEX_NAME: e2e-${{ steps.gen_suffix.outputs.INDEX_NAME_SUFFIX }}
-        PINECONE_ENVIRONMENT: ${{ matrix.pinecone-plan == 'paid' && secrets.PINECONE_ENVIRONMENT_3 || secrets.PINECONE_ENVIRONMENT_4 }}
-        PINECONE_API_KEY: ${{ matrix.pinecone-plan == 'paid' && secrets.PINECONE_API_KEY_3 || secrets.PINECONE_API_KEY_4 }}
+        PINECONE_API_KEY: ${{ secrets.PINECONE_API_KEY_5 }}
```
igiloh-pinecone, line 85: Let's call it simply `PINECONE_API_KEY`?  
The `3` and `4` were used to indicate "pod" VS "serverless", not versions...

pyproject.toml
```diff
@@ -33,11 +32,14 @@ pinecone-text = "^0.7.1"
 tokenizers = "^0.15.0"
 transformers = "^4.35.2"
 sentencepiece = "^0.1.99"
+pandas = "2.0.0"
+pyarrow = "^14.0.1"
```
igiloh-pinecone, line 36: Why do we need these? Are they coming from the client?  
Is there any code in Canopy that directly imports `pandas` or `pyarrow`? If not I think we can remove from here, and rely on the Client's dependencies indirectly.

igiloh-pinecone: I think these were actually coming from `pinecone-datasets` which we removed anyway. Not sure we need them anymore

src/canopy/knowledge_base/knowledge_base.py
```diff
@@ -26,45 +22,46 @@
...
     """
+    pinecone_client = pinecone_client or _get_global_client()
+
+    try:
+        indexes = pinecone_client.list_indexes()
+    except Exception as e:
+        raise RuntimeError("Failed to connect to Pinecone. "
+                           "Please check your credentials and try again") from e
+
+    index_names = [_['name'] for _ in indexes.index_list['indexes']]
```
igiloh-pinecone, line 62:
```suggestion
    index_names = indexes.names()
```
Apparently there's a nice method for this

src/canopy/knowledge_base/knowledge_base.py
```diff
@@ -180,30 +179,35 @@ def __init__(self,
...
 
-    def _connect_index(self,
-                       connect_pinecone: bool = True
-                       ) -> None:
-        if connect_pinecone:
-            connect_to_pinecone()
+        self._default_spec = ServerlessSpec(
+            cloud="aws",
+            region="us-west-2"
+        )
```
igiloh-pinecone, line 199: Why do we need to save this as `self.` member variable?  
You can just create it on the fly in `create_canopy_index()` if `spec is None`, can't you?

src/canopy/knowledge_base/knowledge_base.py
```diff
@@ -668,17 +639,13 @@ def from_config(cls,
...
         return kb
 
-    @staticmethod
-    def _is_starter_env():
-        starter_env_suffixes = ("starter", "stage-gcp-0")
-        return os.getenv("PINECONE_ENVIRONMENT").lower().endswith(starter_env_suffixes)
+    @lru_cache(maxsize=1)
+    def _is_serverless_env(self):
+        description = self._pinecone_client.describe_index(self.index_name)
+        return "serverless" in description["spec"]
```
igiloh-pinecone, line 649: @izellevy `KnowledgeBase` has a dedicated `from_config()` constructor, which does some stuff with `index_params`.  That part should be removed.  

(In the near future we should probably remove this method altogether, but that's a different story..)

igiloh-pinecone (approved):
See a few minor comments

pyproject.toml
izellevy: We need both the dependencies. It should have been a requirement in the first place. So after removing datasets it started to shout so I added them.

src/canopy/knowledge_base/knowledge_base.py
izellevy: Yeah I removed that part.

src/canopy/knowledge_base/knowledge_base.py
izellevy: I can actually let me do it

izellevy: Done

src/canopy/knowledge_base/knowledge_base.py
izellevy: Nice they added recently I guess will change to it

.github/workflows/PR.yml
izellevy: DONE

PR #248, opened 2024-01-15
Temporarily allow tests to fail on starter env

igiloh-pinecone:
We are having problems with this env. Disabling tests for now until we get it figured out

## Type of Change
- [x] Infrastructure change (CI configs, etc)


igiloh-pinecone:
@izellevy merging. Fingers crossed...

PR #258, opened 2024-01-22
Move chunk id generation to base

izellevy:
## Problem

Every chunker has its own way of creating chunk ids. We should generalise that to prevent discrepancies.

## Solution

Created a function in the base and changed all child classes to call that function instead.

## Type of Change

## Test Plan


igiloh-pinecone (approved):
LGTM!

PR #259, opened 2024-01-22
Replace startup with lifespan

izellevy:
## Problem

FastAPI deprecated startup events in favor of lifespan functions.

We are receiving the following warning:

     DeprecationWarning: 
              on_event is deprecated, use lifespan event handlers instead.
      
              Read more about it in the
              [FastAPI docs for Lifespan Events](https://fastapi.tiangolo.com/advanced/events/).
              
        @app.on_event("startup")

## Solution

Replaced on_event with a lifespan function. Removed the shutdown mechanism.

## Type of Change

- [x] Infrastructure change (CI configs, etc)

## Test Plan

No need for an additional test.


igiloh-pinecone:
@izellevy have you tested that this change also works well with the `--reload` option, and with our `/shutdown` mechanism?  
(The latter should be verified on Windows as well, as the behavior there is a bit different)

src/canopy_cli/cli.py
```diff
@@ -666,55 +664,6 @@ def start(host: str, port: str, reload: bool, stream: bool,
...
-        Stop the Canopy server.
-
-        This command sends a shutdown request to the Canopy server.
-        """
-    )
-)
-@click.option("url", "--url", default=DEFAULT_SERVER_URL,
-              help=("URL of the Canopy server to use. "
-                    f"Defaults to {DEFAULT_SERVER_URL}"))
-def stop(url):
```
igiloh-pinecone, line 682: Need to remove from `README` as well

igiloh-pinecone (approved):
@izellevy all in all LGTM.

See one important comment about `canopy stop`

PR #199, opened 2023-12-05
add JinaRecordEncoder

JoanFM:
## Problem

Depends on `pinecone-text` feature to add JinaEncoder https://github.com/pinecone-io/pinecone-text/pull/59. It aims to add JinaRecordEncoder class to allow users to use Jina Embeddings using Jina Embedding API https://jina.ai/embeddings/

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Infrastructure change (CI configs, etc)
- [ ] Non-code change (docs, etc)
- [ ] None of the above: (explain here)

PR #263, opened 2024-01-23
Add `SentenceTransformersRecordEncoder`

tomaarsen:
Hello!

## Problem
This PR ports the `SentenceTransformerEncoder` from `pinecone_text` into `canopy`, allowing various Sentence Transformer models to be used. For example, most of [these models](https://huggingface.co/models?library=sentence-transformers) should now be usable.

## Solution

I based my work here on [`pinecone_text`](https://github.com/pinecone-io/pinecone-text) & the JinaRecordEncoder PR: #199.

## Type of Change

## Test Plan

You can attempt to run the tests to verify whether the default tiny `all-MiniLM-L6-v2` model can be used.

- Tom Aarsen


tomaarsen:
Hello!

First of all, I've ran flake8 and resolved the issues to satisfy your linter.

As for your comments: I think the benefit of a merger depends on whether you wish to use `pinecone-text` outside of the context of RAG or Canopy. I'm not very familiar with it, but it seems like a relatively light-weight library for Dense, Sparse and Hybrid embeddings that might see use elsewhere too. From my perspective, `pinecone-text` feels like a building block, and `canopy` an application putting that building block in an end-to-end application.

I think a merger means that you will end up only with an application, and users wouldn't quickly try and dig out the building block from the application for their own uses. That's just my suspicion though, and it's up to you all to decide whether that's even a problem anyways - I don't think `pinecone-text` had that big of an audience at this point after all.

That all said, the complexity is certainly lower if you merge everything. While they are two separate packages, you need to consider breaking changes between various combinations of versions of both projects, and you'll be more restricted as a result. If `canopy` becomes popular but `pinecone_text` does not, then it will be odd to ask users of the popular project to contribute to the lesser known project instead with extra e.g. embedding models.

And then there's also marketing considerations, which I'm not an expert on.

- Tom Aarsen

tomaarsen:
It looks like I should add `pinecone[dense]` as an extra to the `pyproject.toml` [here](https://github.com/pinecone-io/canopy/blob/6ba92c53a40ba4d9d3cec208eb36fdd46f4328f0/pyproject.toml#L41), so it'll be installed on the CI. What should I call the extra?

igiloh-pinecone (changes requested):
@tomaarsen please see a few minor suggestions

igiloh-pinecone:
> It looks like I should add `pinecone[dense]` as an extra to the `pyproject.toml` [here](https://github.com/pinecone-io/canopy/blob/6ba92c53a40ba4d9d3cec208eb36fdd46f4328f0/pyproject.toml#L41), so it'll be installed on the CI. What should I call the extra?

@tomaarsen thanks for your efforts, and again a very good catch!  
I just opened https://github.com/pinecone-io/canopy/pull/266 which would address that need. You can assume it will be merged to `main` very soon.

src/canopy/knowledge_base/record_encoder/sentence_transformers.py
```diff
@@ -0,0 +1,29 @@
...
+        Initialize the SentenceTransformerRecordEncoder
+
+        Args:
+            model_name: The name of the embedding model to use. See https://huggingface.co/models?library=sentence-transformers
+                        for all possible Sentence Transformer models.
+            batch_size: The number of documents or queries to encode at once.
+                        Defaults to 400.
+            **kwargs: Additional arguments to pass to the underlying `pinecone-text.SentenceTransformerEncoder`.
+        """  # noqa: E501
+        encoder = SentenceTransformerEncoder(model_name, **kwargs)
```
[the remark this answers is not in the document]
tomaarsen: Implemented via 2f821e4a61dfb5c5dc10129f0f500f202d06069d & 92d13e51484b75aa22a1eb3deaccdd1f847f9f76, let me know if you would like `model_name` to be renamed to `document_encoder_name`. Personally, I think `model_name` is more intuitive, but that's the phrasing that I'm familiar with.

tests/system/record_encoder/test_sentence_transformers_encoder.py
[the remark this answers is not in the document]
tomaarsen: Skipped these tests in 81f21331eae2a80fa64925c88c17e15859b1fa87!

src/canopy/knowledge_base/record_encoder/sentence_transformers.py
```diff
@@ -0,0 +1,44 @@
...
+                        See https://huggingface.co/models?library=sentence-transformers
+                        for all possible Sentence Transformer models.
+                        Defaults to `model_name`.
+            batch_size: The number of documents or queries to encode at once.
+                        Defaults to 400.
+            device: The local device to use for encoding, for example "cpu", "cuda" or "mps".
+                        Defaults to "cuda" if cuda is available, otherwise to "cpu".
+            **kwargs: Additional arguments to pass to the underlying `pinecone-text.SentenceTransformerEncoder`.
+        """  # noqa: E501
+        encoder = SentenceTransformerEncoder(
```
igiloh-pinecone, line 38: We should probably wrap this in a `try`-`except` clause, and handle typical HF errors, re-raising with a clear and actionable error message.

src/canopy/knowledge_base/record_encoder/sentence_transformers.py
```diff
@@ -0,0 +1,44 @@
...
+            document_encoder_name=model_name,
+            query_encoder_name=query_encoder_name,
+            device=device,
+            **kwargs,
+        )
+        super().__init__(dense_encoder=encoder, batch_size=batch_size)
```
igiloh-pinecone, line 44: Please consider implementing the `_format_error(self, err)` method, which gets called if there was a runtime error at encoding time.  
We try to anticipate the typical misconfigurations, and re-raise these with a clear and actionable error message

src/canopy/knowledge_base/record_encoder/sentence_transformers.py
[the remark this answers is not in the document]
igiloh-pinecone: I agree that `model_name` is more intuitive. We'll consider a deeper renaming in the near future.  
Thanks!

igiloh-pinecone:
@tomaarsen thank you for your very quick response!  
Please see two comments about error handling (I added them a few minutes after publishing the official review). Other than that LGTM and I think we can approve and merge

tomaarsen:
I'll have a look at them tomorrow :)

- Tom Aarsen

tomaarsen:
Apologies for the delay. I've wrapped the __init__ in a try-except. To my knowledge, there are two common errors here: `ImportError` and `RepositoryNotFoundError`. `pinecone-text` already has a great actionable `ImportError` for the former, so I won't override that one. 

As for `_format_error(self, err)` - SentenceTransformers does not raise any specific errors for misconfigurations on runtime. It also does not have e.g. RateLimit exceptions and such, so I'm unsure what I could reasonably implement here.

Would you like additional changes?

- Tom Aarsen

igiloh-pinecone:
> Apologies for the delay. I've wrapped the `__init__` in a try-except. To my knowledge, there are two common errors here: `ImportError` and `RepositoryNotFoundError`. `pinecone-text` already has a great actionable `ImportError` for the former, so I won't override that one.

Actually we should probably wrap the `ImportError`  from `pinecone-text` asking users to install `pinecone-text[dense]` and replace it with a message to install `canopy-sdk[torch]` instead, right? (see https://github.com/pinecone-io/canopy/pull/266) 

> As for `_format_error(self, err)` - SentenceTransformers does not raise any specific errors for misconfigurations on runtime. It also does not have e.g. RateLimit exceptions and such, so I'm unsure what I could reasonably implement here.

I was thinking about stuff like catching OOM errors and asking user to change the `batch_size` config param, or other such cases where the actionable fix is related to Canopy's config.   
But if there aren't any (or they are not too obvious) - sure, let's keep it as is.

> Would you like additional changes?

No, thank you very much for your efforts!  
Approving the PR now

> Edit: How would you like me to counteract the mypy failure?

Given that it looks like a [known issue](https://github.com/huggingface/huggingface_hub/issues/1662), I guess there's not much we can do about it other than adding `'huggingface_hub.utils'` to mypy's override list in the `pyproject.toml` file.

igiloh-pinecone (approved):
LGTM!

tests/system/record_encoder/test_sentence_transformers_encoder.py
```diff
@@ -0,0 +1,61 @@
...
+           Query(text="Sample query 2"),
+           Query(text="Sample query 3"),
+           Query(text="Sample query 4")]
+
+
+@pytest.fixture
+def encoder():
+    try:
+        encoder = SentenceTransformerRecordEncoder(batch_size=2)
+    except ImportError as e:
```
tomaarsen, line 29:
```suggestion
    except ImportError:
```
`e` was unused, this should hopefully satisfy mypy.

tests/system/record_encoder/test_sentence_transformers_encoder.py
```diff
@@ -0,0 +1,61 @@
...
+        pytest.skip(
+            f"`transformers` extra not installed. Skipping SentenceTransformer system "
+            f"tests"
```
tomaarsen, line 32:
```suggestion
            "`transformers` extra not installed. Skipping SentenceTransformer system "
            "tests"
```
The `f` prefix should not be used here

tests/unit/record_encoder/test_sentence_transformers_encoder.py
```diff
@@ -0,0 +1,74 @@
...
+           Query(text="Sample query 2"),
+           Query(text="Sample query 3"),
+           Query(text="Sample query 4")]
+
+
+@pytest.fixture
+def encoder():
+    try:
+        encoder = SentenceTransformerRecordEncoder(batch_size=2)
+    except ImportError as e:
```
tomaarsen, line 31:
```suggestion
    except ImportError:
```
idem

tests/unit/record_encoder/test_sentence_transformers_encoder.py
```diff
@@ -0,0 +1,74 @@
...
+        pytest.skip(
+            f"`transformers` extra not installed. Skipping SentenceTransformer unit "
+            f"tests"
```
tomaarsen, line 34:
```suggestion
            "`transformers` extra not installed. Skipping SentenceTransformer unit "
            "tests"
```
idem

src/canopy/knowledge_base/record_encoder/sentence_transformers.py
```diff
@@ -0,0 +1,57 @@
...
+            raise RuntimeError(
+                "Your chosen Sentence Transformer model(s) could not be found. "
+                f"Details: {str(e)}"
+            ) from e
+        except ImportError as e:
+            raise ImportError(
+                f"{self.__class__.__name__} requires the `torch` and `transformers` "
+                f"extra dependencies. Please install them using "
+                f"`pip install canopy-sdk[torch,transformers]`."
+            )
```
tomaarsen, line 56:
```suggestion
            ) from e
```
Satisfies mypy & correctly links the errors.

tomaarsen (commented):
A few comments on the recent changes :)

PR #265, opened 2024-01-23
Add HybridRecordEncoder

izellevy:
## Problem

We don't have support for [Hybrid Search](https://www.pinecone.io/learn/hybrid-search-intro/) in canopy.

## Solution

Adds a new HybridRecordEncoder that uses the given dense encoder and the BM25 sparse encoder for
hybrid search purposes.

## Type of Change

## Test Plan

Added relevant tests.


src/canopy/knowledge_base/record_encoder/hybrid.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        Initialize the encoder.
+
+        Args:
+            dense_encoder: A BaseDenseEncoder to encode the text.
+            alpha: The weight of the dense vector in the hybrid representation (between 0 and 1).
+            **kwargs: Additional arguments to pass to the RecordEncoder.
+        """  # noqa: E501
+
+        if not 0 <= alpha <= 1:
```
acatav, line 42: In V4 alpha cannot be zero because in boosting the user would get kind of random results, so I suggest to change condition to `0 < alpha <=1`

tests/unit/record_encoder/test_hybrid_record_encoder.py
```diff
@@ -0,0 +1,92 @@
...
+        HybridRecordEncoder(inner_encoder, alpha=-1)
+    with pytest.raises(ValueError):
+        HybridRecordEncoder(inner_encoder, alpha=2)
+
+
+def test_encode_documents(encoder, documents, queries):
+    encoded_documents = encoder.encode_documents(documents)
+    for encoded_document in encoded_documents:
+        assert len(encoded_document.values) == encoder.dimension
+        assert len(encoded_document.sparse_values) > 0
```
acatav, line 58: I suggest to at least assert that the format of the sparse values is correct (a dict with the expected keys)

tests/unit/record_encoder/test_hybrid_record_encoder.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def test_encode_queries(encoder, queries):
+    encoded_queries = encoder.encode_queries(queries)
+    assert len(encoded_queries) == len(queries)
```
acatav, line 63: I suggest to also here test that there are some sparse values and that they are in the right format

src/canopy/knowledge_base/record_encoder/hybrid.py
```diff
@@ -0,0 +1,111 @@
...
+            raise ValueError("Alpha must be between 0 and 1")
+
+        super().__init__(**kwargs)
+        self._dense_encoder = dense_encoder
+        self._sparse_encoder = BM25Encoder.default()
```
acatav, line 47: IMO we should support loading a user file from day 1 because this is what we recommend also to users. It simply adding an optional file path and call `load` if it's not none

tests/unit/record_encoder/test_hybrid_record_encoder.py
acatav, line 63: I suggest to test that alpha was applied correctly

src/canopy/knowledge_base/record_encoder/hybrid.py
```diff
@@ -0,0 +1,111 @@
...
+    Alpha is a parameter that controls the weight of the dense vector in the hybrid representation.
+    If alpha is 1, the query vector will be the dense vector, if alpha is 0, the query vector will be the sparse vector.
+    The default value of alpha is 0.5.
+
+    For more information about the encoders see: https://github.com/pinecone-io/pinecone-text
+
+    """  # noqa: E501
+
+    def __init__(self,
+                 dense_encoder: BaseDenseEncoder,
```
acatav, line 30: I wonder if this should have default if none of OpenAI encoder, not sure

acatav (approved):
LGTM, see my small comments

src/canopy/knowledge_base/record_encoder/hybrid.py
acatav: We should raise error for alpha=0 saying that for V4 it can't be used, and that for both v3 the user can simply pass a very small alpha and get very similar results to alpha=0

src/canopy/knowledge_base/record_encoder/hybrid.py
```diff
@@ -0,0 +1,122 @@
...
+    def dimension(self) -> int:
+        """
+        The dimension is the length of the vector generated by the `DenseRecordEncoder`
+        Canopy will run a single word through the encoder to get the dimension, this will also validate that the encoder
+        is working properly.
+
+        Returns:
+            dimension(int): the dimension of the encoder
+        """  # noqa: E501
+        dummy_doc = KBDocChunk(text="hello", id="dummy_doc", document_id="dummy_doc")
```
igiloh-pinecone, line 113: You can simply call `return self.dense_encoder.dimension`.  
That's better, since in some of the dense encoders we wrap that call with proper error handling etc. Not a good idea to re-implement here.

src/canopy/models/data_models.py
```diff
@@ -3,10 +3,15 @@
 from typing import Optional, List, Union, Dict, Literal
 
 from pydantic import BaseModel, Field, validator, Extra
+from typing import TypedDict
 
 Metadata = Dict[str, Union[str, int, float, List[str]]]
 
 
+class SparseVector(TypedDict):
```
igiloh-pinecone, line 11: Any reason why not to import `pinecone-text`'s definition of `SparseVector`, instead of duplicating it?

tests/unit/record_encoder/test_hybrid_record_encoder.py
izellevy: Done

tests/unit/record_encoder/test_hybrid_record_encoder.py
izellevy: Done

src/canopy/knowledge_base/record_encoder/hybrid.py
izellevy: Done

tests/unit/record_encoder/test_hybrid_record_encoder.py
izellevy: Done

src/canopy/knowledge_base/record_encoder/hybrid.py
izellevy: Done

src/canopy/models/data_models.py
izellevy: Yep the pinecone-text definition yields to an incorrect parsing on KBEncodedDocChunk. Basically Pydantic casts Pinecone Text's SparseVector incorrectly causing sparse indices to be float instead of integers.(definition of SparseVector = Dict[str, Union[List[int], List[float]]]) Pydantic in that cases parses to the latter (List[float]). 
 
In the new serverless floats are not allowed as indices which raises an exception. In any case we should aim to fix that on pinecone-text because this typing is incorrect now. Better to use typed dict for these purposes. For now I defined in canopy to avoid waiting for a new release.

src/canopy/knowledge_base/record_encoder/hybrid.py
```diff
@@ -0,0 +1,122 @@
+from typing import List
+from functools import cached_property
+
+from pinecone_text.dense import OpenAIEncoder
+from pinecone_text.dense.base_dense_ecoder import BaseDenseEncoder
+from pinecone_text.hybrid import hybrid_convex_scale
+from pinecone_text.sparse import BM25Encoder
```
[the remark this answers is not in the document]
izellevy: I agree in general if we assume pinecone_text.BaseDenseEncoder & pinecone_text.BM25Encoder exists. I tried replacing but it causes super long paths that are duplicating making the code much less readable. The modules are not exposed properly. I would suggest to wait for a clean API (if we will keep supporting pinecone-text) before changing it to pinecone-text.XXX I think in any case we will move pinecone-text inside the library, then it might be much easier to make some changes.

src/canopy/knowledge_base/record_encoder/hybrid.py
izellevy: Done

src/canopy/knowledge_base/record_encoder/hybrid.py
```diff
@@ -0,0 +1,125 @@
...
+
+    Alpha is a parameter that controls the weight of the dense vector in the hybrid representation.
+    If alpha is 1, the query vector will be the dense vector. The default value of alpha is 0.5.
+
+    For more information about the encoders see: https://github.com/pinecone-io/pinecone-text
+
+    """  # noqa: E501
+
+    def __init__(self,
+                 dense_encoder: BaseDenseEncoder = None,
```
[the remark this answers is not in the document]
izellevy: Changed it to DenseRecordEncoder

src/canopy/knowledge_base/record_encoder/hybrid.py
```diff
@@ -0,0 +1,136 @@
...
+        Args:
+            dense_record_encoder: A DenseRecordEncoder to encode the text.
+            alpha: The weight of the dense vector in the hybrid representation (between 0 and 1).
+            bm_25_encoder_df_path: The path to the file that contains the document frequencies of the BM25Encoder.\
+            You can create this file by fitting the BM25Encoder on a corpus of documents and calling `dump`\
+            on the encoder.
+            **kwargs: Additional arguments to pass to the RecordEncoder.
+        """  # noqa: E501
+
+        if not 0 < alpha <= 1:
```
igiloh-pinecone, line 54: We should probably add an explicit error message for `alpha==0`, saying "sparse only not supported"

tests/unit/record_encoder/test_hybrid_record_encoder.py
```diff
@@ -0,0 +1,119 @@
...
+def test_encode_queries(hybrid_encoder, queries):
+    encoded_queries = hybrid_encoder.encode_queries(queries)
+    assert len(encoded_queries) == len(queries)
+    for encoded_query in encoded_queries:
+        assert len(encoded_query.values) == hybrid_encoder.dimension
+        assert "indices" in encoded_query.sparse_values
+        assert "values" in encoded_query.sparse_values
+
+
+def test_encode_queries_alpha_applied_correctly(hybrid_encoder, queries):
```
igiloh-pinecone, line 89: @izellevy I don't think this test is really testing what you're saying it tests.  

1. It's not testing that changing `alpha` as init parameter really works, since you patched over it.
2. It's not really checking that the output is `alpha * dense_value`, just that if you changed `alpha` the output also changes
3. It doesn't verify that sparse is affected by `(1 - alpha)`.

I suggest testing this feature exactly opposite: Create a new `HybridRecrordEncoder` and pass a deliberate `alpha` value to the contructor (e.g. `0.2`). Then, path the `encoder._dense_encoder.encode_queries()` and `encoder._sparse_encoder.encode_queries()` methods, controlling what they will output.  
Now you can check that calling `encoder.encode_queries()` really returns a value which is what expected.

tests/unit/record_encoder/test_hybrid_record_encoder.py
```diff
@@ -0,0 +1,119 @@
...
+    assert hybrid_encoder.dimension == inner_dimension
+
+
+def test_init_encoder_invalid_alpha(dense_record_encoder):
+    with pytest.raises(ValueError):
+        HybridRecordEncoder(dense_record_encoder, alpha=-1)
+    with pytest.raises(ValueError):
+        HybridRecordEncoder(dense_record_encoder, alpha=2)
+    with pytest.raises(ValueError):
+        HybridRecordEncoder(dense_record_encoder, alpha=0)
```
igiloh-pinecone, line 69:
```suggestion
    with pytest.raises(ValueError, match="sparse only):
        HybridRecordEncoder(dense_record_encoder, alpha=0)
```
See suggestion above

igiloh-pinecone (commented):
@izellevy see a few last comments about unit tests

tests/unit/record_encoder/test_hybrid_record_encoder.py
izellevy: 1) I now added created RecordEncoder with alpha
2) Right but then I kinda duplicate this test: https://github.com/pinecone-io/pinecone-text/blob/7dff87b1262f8b22df5c98ebd18528329e107b62/tests/unit/test_hybrid_convex_scale.py#L11
3) Now I also check that the scaling applied correctly to sparse.

If the sparse and dense matches with scaling I think we can assume it is calculated correctly.

src/canopy/knowledge_base/record_encoder/hybrid.py
izellevy: Added

tests/unit/record_encoder/test_hybrid_record_encoder.py
izellevy: Done

tests/unit/record_encoder/test_hybrid_record_encoder.py
igiloh-pinecone: OK...

igiloh-pinecone (approved):
LGTM

PR #266, opened 2024-01-24
Make transformers optional + allow pinecone-text[dense]

igiloh-pinecone:
## Problem

1. `transformers` shouldn't be a required dependency. It was accidentally added as such.
2. No way to install `pinecone-text` with its `[dense]` extra dependency

## Solution

1. Made `transformers` optional, only installed with a new `anyscale` extra
2. Added a new `torch` extra (torch is the heaviest installation, so it made more sense as the extra's name than `transformers` or `sentence-transformers`, which might be ambiguous)

## Type of Change

Might be slightly breaking for users who currently use the Anyscale Tokenizer, and would need to re-install with the extra.

## Test Plan

No affect on current tests

### DIFF
--- a/.github/actions/install-deps-and-canopy/action.yml
+++ b/.github/actions/install-deps-and-canopy/action.yml
@@ -43,7 +43,7 @@ runs:
     if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
     env:
       CANOPY_EXTRAS: ${{ inputs.extras }}
-    run: poetry install --no-interaction --no-root --with dev --extras "{{ env.CANOPY_EXTRAS }}"
+    run: poetry install --no-interaction --no-root --with dev --extras "{$CANOPY_EXTRAS}"
   - name: Install project
     if: ${{ inputs.install-canopy == 'true' }}
     shell: bash
--- a/.github/actions/install-deps-and-canopy/action.yml
+++ b/.github/actions/install-deps-and-canopy/action.yml
@@ -43,7 +43,7 @@ runs:
     if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
     env:
       CANOPY_EXTRAS: ${{ inputs.extras }}
-    run: poetry install --no-interaction --no-root --with dev --extras "{$CANOPY_EXTRAS}"
+    run: poetry install --no-interaction --no-root --with dev --extras "${CANOPY_EXTRAS}"
   - name: Install project
     if: ${{ inputs.install-canopy == 'true' }}
     shell: bash
--- a/.github/actions/install-deps-and-canopy/action.yml
+++ b/.github/actions/install-deps-and-canopy/action.yml
@@ -9,9 +9,15 @@ inputs:
     description: "Whether to install canopy library, or dependencies only"
     required: true
     default: "true"
+  extras:
+    description: "Extra dependencies to install"
+    required: false
+    default: "cohere transformers"
 
 runs:
   using: "composite"
+  env:
+    CANOPY_EXTRAS: ${{ inputs.extras }}
   steps:
   - name: Install Poetry
     uses: snok/install-poetry@v1
@@ -37,8 +43,8 @@ runs:
   - name: Install dependencies
     shell: bash
     if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
-    run: poetry install --no-interaction --no-root --all-extras --with dev
+    run: poetry install --no-interaction --no-root --with dev --extras "{{ env.CANOPY_EXTRAS }}"
   - name: Install project
     if: ${{ inputs.install-canopy == 'true' }}
     shell: bash
-    run: poetry install --no-interaction --all-extras --with dev
+    run: make install-extras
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -28,16 +28,20 @@ types-pyyaml = "^6.0.12.12"
 jsonschema = "^4.2.0"
 types-jsonschema = "^4.2.0"
 prompt-toolkit = "^3.0.39"
-pinecone-text = "^0.7.2"
+pinecone-text = [{version = "^0.7.2"},
```
igiloh-pinecone, line 31: Found this syntax [here](https://github.com/python-poetry/poetry/issues/5733#issuecomment-1151481078). 
Was skeptical at first - but it seems to be working

pyproject.toml
```diff
@@ -28,16 +28,20 @@ types-pyyaml = "^6.0.12.12"
 jsonschema = "^4.2.0"
 types-jsonschema = "^4.2.0"
 prompt-toolkit = "^3.0.39"
-pinecone-text = "^0.7.2"
```
igiloh-pinecone, line 31: Found this syntax [here](https://github.com/python-poetry/poetry/issues/5733#issuecomment-1151481078). 
Was skeptical at first - but it seems to be working

pyproject.toml
```diff
@@ -28,16 +28,20 @@ types-pyyaml = "^6.0.12.12"
...
-transformers = "^4.35.2"
+transformers = {version = "^4.35.2", optional = true}
 sentencepiece = "^0.1.99"
 pandas = "2.0.0"
 pyarrow = "^14.0.1"
 cohere = { version = ">=4.37", optional = true }
 
 [tool.poetry.extras]
 cohere = ["cohere"]
+torch = ["pinecone-text"]
```
acatav, line 43: Are you sure this could be `torch` and not have to be `dense`? from the link you sent the used the extras name but maybe it's just in the example and not a must

pyproject.toml
```diff
@@ -28,16 +28,20 @@ types-pyyaml = "^6.0.12.12"
 jsonschema = "^4.2.0"
 types-jsonschema = "^4.2.0"
 prompt-toolkit = "^3.0.39"
-pinecone-text = "^0.7.2"
+pinecone-text = [{version = "^0.7.2"},
+                 {version = "^0.7.2", extras = ["dense"], optional = true}]
+
 tokenizers = "^0.15.0"
-transformers = "^4.35.2"
+transformers = {version = "^4.35.2", optional = true}
```
acatav, line 35: Can we remove this line? seems like it will be taken from pinecone text anyway

acatav (changes requested):
added few questions

pyproject.toml
igiloh-pinecone: We need it for `LLamaTokenizer`, which is used with Anyscale's `llama2` LLM. That's why I added an `anyscale` extra.

pyproject.toml
igiloh-pinecone: It can be anything you want, it's just a name (I tested it).  
When you `poetry install -E torch` it's interpreted as "the `optional` version of `pinecone-text`, as defined above".

.github/actions/install-deps-and-canopy/action.yml
```diff
@@ -9,6 +9,10 @@ inputs:
     description: "Whether to install canopy library, or dependencies only"
     required: true
     default: "true"
+  extras:
+    description: "Extra dependencies to install"
```
izellevy, line 13: Maybe we should explain it is space separated

izellevy: Taking a list also is an option

.github/actions/install-deps-and-canopy/action.yml
```diff
@@ -37,8 +41,12 @@ runs:
   - name: Install dependencies
     shell: bash
     if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
-    run: poetry install --no-interaction --no-root --all-extras --with dev
+    env:
```
izellevy, line 44: Do we need --with dev? By default it is with dev no?

igiloh-pinecone: I wasn't sure, and I prefer to be explicit...

.github/actions/install-deps-and-canopy/action.yml
igiloh-pinecone, line 13:
```suggestion
    description: "Extra dependencies to install, space separated"
```

src/canopy_cli/cli.py
```diff
@@ -338,7 +339,7 @@ def upsert(index_name: str,
 
     try:
         kb.connect()
-    except RuntimeError as e:
+    except (RuntimeError) as e:
```
izellevy, line 342: Do we need parantheses?

izellevy (approved):

src/canopy_cli/cli.py
igiloh-pinecone: Not really...
```suggestion
    except RuntimeError as e:
```

PR #267, opened 2024-01-25
Updates to README to be more accurate.

aulorbe:
## Problem

Quick updates to README to reflect current state of things.

## Solution

Describe the approach you took. Link to any relevant bugs, issues, docs, or other resources.

## Type of Change

- [x] Non-code change (docs, etc)

## Test Plan

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,6 +258,8 @@ To run the canopy server for production, please run:
 gunicorn canopy_server.app:app --worker-class uvicorn.workers.UvicornWorker --bind 0.0.0.0:PORT --workers WORKER_COUNT
 ```
 
+Note: replace `PORT` and `WORKER_COUNT` with values.
+
 Alternatively, consider utilizing the Canopy Docker image available on [GitHub Packages](https://github.com/pinecone-io/canopy/pkgs/container/canopy) 
 for your production needs. For guidance on deploying Canopy on the Google Cloud Platform (GCP), refer to the example provided in the
 [Deployment to GCP](docs/deployment-gcp.md) documentation.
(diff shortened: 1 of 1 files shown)


aulorbe:
@ilshatahatov Review?

README.md
```diff
@@ -258,6 +258,8 @@ To run the canopy server for production, please run:
 gunicorn canopy_server.app:app --worker-class uvicorn.workers.UvicornWorker --bind 0.0.0.0:PORT --workers WORKER_COUNT
 ```
 
+Note: replace `PORT` and `WORKER_COUNT` with values.
```
igiloh-pinecone, line 261: No, no, that's actually a typo 🤦 .  
This was supposed to be `$PORT` and `$WORKER_COUNT`. And I guess these come from our docker image, where we explicitly set these env vars (with default, even if the user hasn't provided them).  
Can you fix the typo please?

igiloh-pinecone: Actually, I'm not even sure why @miararoy has changed the main README to use these env vars - unlike in the docker case, if someone just `pip install canopy`, he is **not** guaranteed to have those env vars at all. I think we should just revert to what we had [before](https://github.com/pinecone-io/canopy/commit/44cfad11add34fc61d6e9f9f0ac4071ff24195bf).

@miararoy ?

igiloh-pinecone (approved):
@aulorbe all in all LGTM.

Please see my comment about that `PORT` and `WORKER_COUNT` thing. I think we should just go back to the original phrasing, and then we don't need more explanations later.

PR #273, opened 2024-02-04
Add dimension support for OpenAI embeddings

izellevy:
## Problem

New OpenAI embedding models support controlling the dimension parameter for the embeddings vectors. We cannot currently pass a dynamic dimension to OpenAI models.

## Solution

Updated the pinecone-text package version that supports passing dimension to the Encoder. Changed the default embedding model to the new one (text-embedding-3-small). Removed support for taking the dimension on create_canopy_index. From now on, we always rely on the RecordEncoder's dimension to create the index.

## Type of Change

## Test Plan


acatav (approved):
LGTM!

PR #277, opened 2024-02-06
Change the default poetry extras

izellevy:
## Problem

By default we install all-extras in dockerfile that causes out of disk error while building the image. We need to shrink the image size.

## Solution

Now we install only the necessary extras that dropped the image size from 6GB to 600MB.

## Type of Change

## Test Plan

### DIFF
--- a/.github/actions/install-deps-and-canopy/action.yml
+++ b/.github/actions/install-deps-and-canopy/action.yml
@@ -9,11 +9,6 @@ inputs:
     description: "Whether to install canopy library, or dependencies only"
     required: true
     default: "true"
-  extras:
-    description: "Extra dependencies to install, space separated"
-    required: false
-    default: "cohere transformers"
-
 runs:
   using: "composite"
   steps:
@@ -41,12 +36,8 @@ runs:
   - name: Install dependencies
     shell: bash
     if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
-    env:
-      CANOPY_EXTRAS: ${{ inputs.extras }}
-    run: poetry install --no-interaction --no-root --with dev --extras "${CANOPY_EXTRAS}"
+    run: make install POETRY_INSTALL_EXTRA_ARGS="--no-interaction --no-root --with dev"
   - name: Install project
     if: ${{ inputs.install-canopy == 'true' }}
     shell: bash
-    env:
-      CANOPY_EXTRAS: ${{ inputs.extras }}
-    run: make install-extras
+    run: make install POETRY_INSTALL_EXTRA_ARGS="--with dev"
--- a/.github/actions/install-deps-and-canopy/action.yml
+++ b/.github/actions/install-deps-and-canopy/action.yml
@@ -36,8 +36,8 @@ runs:
   - name: Install dependencies
     shell: bash
     if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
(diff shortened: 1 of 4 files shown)


.github/actions/install-deps-and-canopy/action.yml
```diff
@@ -41,12 +36,8 @@ runs:
...
-      CANOPY_EXTRAS: ${{ inputs.extras }}
-    run: poetry install --no-interaction --no-root --with dev --extras "${CANOPY_EXTRAS}"
+    run: make install POETRY_INSTALL_EXTRA_ARGS="--no-interaction --no-root --with dev"
   - name: Install project
     if: ${{ inputs.install-canopy == 'true' }}
     shell: bash
-    env:
-      CANOPY_EXTRAS: ${{ inputs.extras }}
-    run: make install-extras
+    run: make install POETRY_INSTALL_EXTRA_ARGS="--with dev"
```
igiloh-pinecone, line 43: We should `--no-interaction` here as well

Makefile
```diff
@@ -1,20 +1,24 @@
 TEST_WORKER_COUNT = 8
 
-REPOSITORY = ghcr.io/pinecone-io/canopy
+POETRY_INSTALL_SYSTEM_ARGS = -E cohere -E transformers
+POETRY_INSTALL_EXTRA_ARGS =
```
igiloh-pinecone, line 4: The names here are really confusing, with `SYSTEM_ARGS` being a list of `extra`s, and `EXTRA_ARGS` being used for passing system arguments like `--no-root`...

Let's use something like `DEFAULT_EXTRAS` and `POETRY_INSTALL_ARGS`.

Makefile
```diff
@@ -23,10 +27,7 @@ static:
 	poetry run mypy src
 
 install:
-	poetry install
```
igiloh-pinecone, line 26: That's not the right approach. The default behavior for most users would still be to install canopy itself only.  

The `install-extras` is meant for our CI workflows, which we want to install all the "default" extras on top of base canopy (but not --all-extras`).

Makefile
```diff
@@ -42,34 +43,41 @@ test-e2e:
...
 
 docker-run-dev:
-	docker run $(CONTAINER_COMMON_RUN_ARGS) $(CONTAINER_EXTRA_RUN_ARGS) -it $(REPOSITORY)-dev:$(IMAGE_TAG)
+	docker run $(CONTAINER_SYSTEM_RUN_ARGS) $(CONTAINER_EXTRA_RUN_ARGS) -it $(REPOSITORY)-dev:$(IMAGE_TAG)
+
+print-var:
+	@echo $($(VAR))
 
 help:
 	@echo "Available targets:"
```
igiloh-pinecone, line 64: We should consider some pattern for automatically generating this help message by adding "docstrings" to the commands themselves.  
There are a few clever `Makefile` tricks that allow that (with some `awk` or `sed` or something).  

Not for this PR, just a general thought for the future

igiloh-pinecone (changes requested):
@izellevy please see a few comments

.github/actions/install-deps-and-canopy/action.yml
izellevy: Done

Makefile
izellevy: Changed it

acatav (approved):

PR #278, opened 2024-02-07
Move config dir to be part of the lib

acatav:
## Problem

Right now there is no way to access the config dir that compatible with the installation

## Solution

Make the config dir part of the library

## Type of Change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Test Plan

Adjusted the relevant test to work with the new config path


igiloh-pinecone (dismissed):
@acatav two comments:

1. Please merge `main`, there's a new `cohere.yaml` config as well.
2. We need to give users easy access to config templates.  
   Please add a `canopy gen-config PATH` cli command that "generates" config templates by finding them in the installed package and copying to a user's selected path. We should also add this to the README.

izellevy (approved):

src/canopy_cli/cli.py
```diff
@@ -210,6 +211,40 @@ def health(url):
...
+    if not template.endswith('.yaml'):
+        template += '.yaml'
+
+    if os.path.exists(path):
+        click.confirm(click.style(f"File {path} already exists. Overwrite?", fg="red"),
+                      abort=True)
+
+    try:
+        with pkg_resources.open_text('canopy.config', template) as f:
+            config = yaml.safe_load(f)
```
igiloh-pinecone, line 228: @acatav you don't need this tedious loading and saving code. Just copy the entire content of the `config/` dir from the installed location to the user's desired path

PR #287, opened 2024-02-11
Fix create config CLI command

acatav:
## Problem

Current CLi command for creating config templates not working

## Solution

1. Use simple file paths from within the repo to copy config files
2. For ease of use add a util function to write config files

## Type of Change

- [X] New feature (non-breaking change which adds functionality)

## Test Plan

Changed the relevant tests


src/canopy/utils/config.py
```diff
@@ -1,6 +1,18 @@
...
 
 
+def load_config_template(name: str) -> Dict[str, Any]:
+    if not name.endswith(".yaml"):
+        name += ".yaml"
+
+    config_template_path = os.path.join(os.path.dirname(os.path.dirname(__file__)),
+                                        "config_templates", name)
+    with open(config_template_path, "r") as f:
+        return yaml.safe_load(f)
```
izellevy, line 13: Maybe have a Directory class in the init file to refer to specific directories with the project. Sth like:

class Directory:
    Root = Path(....)
    ConfigTemplates = self.Root.joinpath(...)

igiloh-pinecone (approved):
LGTM

PR #288, opened 2024-02-12
Add support for Pydantic v2

izellevy:
## Problem

Currently we only support Pydantic v1 which is a problem for some users.

## Solution

Migrating to Pydantic v2

## Type of Change

## Test Plan


src/canopy/llm/models.py
```diff
@@ -32,11 +32,12 @@ class FunctionParameters(BaseModel):
     required_properties: List[FunctionProperty]
     optional_properties: List[FunctionProperty] = []
 
-    def dict(self, *args, **kwargs):
+    @model_serializer()
```
igiloh-pinecone, line 35: Why did you replace `def dict()` sometimes with `def model_dump` and sometimes with `@model_serializer`?  
Is there a difference between the two? If not, it's probably a good idea to maintain a single code style for all models

tests/unit/chunker/test_markdown_chunker.py
```diff
@@ -127,7 +127,7 @@ def expected_chunks(documents):
                      '\ntext in level 3\n#### Level 4\ntext in level 4\n##### Level 5'
                      '\ntext in level 5\n###### Level 6\ntext in level 6',
                 source='doc_1',
-                metadata={'test': '1'},
```
igiloh-pinecone, line 130: Why was this cast to `str` before? (Seems like we had a bug?)

tests/unit/context_builder/test_stuffing_context_builder.py
```diff
@@ -50,7 +50,7 @@ def setup_method(self):
                         ])
         ]
         self.full_context = Context(
-            content=StuffingContextContent(__root__=[
+            content=StuffingContextContent(root=[
```
igiloh-pinecone, line 53: Again, no need for `root=`

tests/e2e/test_app.py
```diff
@@ -104,15 +104,15 @@ def test_health(client):
     assert health_response.is_success
     assert (
             health_response.json()
-            == HealthStatus(pinecone_status="OK", llm_status="OK").dict()
+            == HealthStatus(pinecone_status="OK", llm_status="OK").model_dump()
```
igiloh-pinecone, line 107: Again, I think that 
```suggestion
            == dict(HealthStatus(pinecone_status="OK", llm_status="OK"))
```
is more readable.  
But totally your call

igiloh-pinecone (commented):
@izellevy see a few suggestions

src/canopy/chat_engine/chat_engine.py
```diff
@@ -221,7 +221,7 @@ def chat(self,
                                                 model_params=model_params_dict)
         debug_info = {}
         if CANOPY_DEBUG_INFO:
-            debug_info['context'] = context.dict()
+            debug_info['context'] = context.model_dump()
```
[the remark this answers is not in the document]
izellevy: Quoting from pydantic docs:
Note that dict(user) will not recursively convert nested models into dicts, but .model_dump() will.

So I think we need to keep it as model_dump

src/canopy/context_engine/context_builder/stuffing.py
```diff
@@ -23,15 +24,12 @@ class ContextQueryResult(BaseModel):
...
-
-    def dict(self, **kwargs):
-        return super().dict(**kwargs)['__root__']
+    root: List[ContextQueryResult]
 
     # In the case of StuffingContextBuilder, we simply want the text representation to
     # be a json. Other ContextContent subclasses may render into text differently
     def to_text(self, **kwargs):
-        return self.json(**kwargs)
+        return json.dumps(self.model_dump(), **kwargs)
```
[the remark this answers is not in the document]
izellevy: Because we pass args to json_dumps (such as indent). This was supported in v1 but not v2 (.model_dump_json does not get indent in pydantic v2)

src/canopy/llm/models.py
izellevy: The only difference is that you cannot pass args and kwargs to model_serializer. In any case I changed to model serializers since we don't use these args for now.

src/canopy/context_engine/context_builder/stuffing.py
```diff
@@ -52,10 +50,10 @@ def build(self,
             ContextQueryResult(query=qr.query, snippets=[])
             for qr in query_results]
         debug_info = {"num_docs": len(sorted_docs_with_origin), "snippet_ids": []}
-        content = StuffingContextContent(__root__=context_query_results)
+        content = StuffingContextContent(root=context_query_results)
```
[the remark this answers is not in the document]
izellevy: Right fixing

tests/unit/chunker/test_markdown_chunker.py
izellevy: Kinda bug of Union parsing in pydantic v1. In pydantic v2 there is smart unions so they fixed this problem.

tests/unit/context_builder/test_stuffing_context_builder.py
izellevy: Fixing

tests/e2e/test_app.py
izellevy: As I said model_dump is recursive, so better to stick onto that.

src/canopy/llm/models.py
izellevy: Actually it failed the tests rolling back I get:

pydantic_core._pydantic_core.PydanticSerializationError: Error calling function `serialize_model`: RecursionError which kinda makes sense

src/canopy/context_engine/context_builder/stuffing.py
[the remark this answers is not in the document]
izellevy: Fixed

igiloh-pinecone (approved):
LGTM

izellevy:
@igiloh-pinecone checked and made sure it works with all extras

Issue #284, opened 2024-02-09
[Bug] CORS preflight headers - OPTIONS method not allowed


coreation:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

When calling the chat completions API endpoint directly from a client like the browser, it makes a preflight CORS check using the OPTIONS HTTP request. This is not supported and results in the API call failing as it does not get past this check.

### Expected Behavior

Unless the API endpoint is solely intended for proxy-like clients that entirely control the HTTP calls and headers made, I would expect the OPTIONS HTTP request being supported. This makes it for browser application clients possible to directly interface with the exposed canopy API.

I believe the exposed API is not protected and if this is done intentionally, I'm glad to close the issue :)

### Steps To Reproduce

1. Run a vanilla set-up of canopy
2. Send an OPTIONS request to the chat/completions endpoint

### Relevant log output

```shell
INFO:     127.0.0.1:55991 - "OPTIONS /v1/qanda-content/chat/completions HTTP/1.1" 405 Method Not Allowed
```

### Environment

```markdown
n/a
```

PR #319, opened 2024-03-11
Issue #284 Add CORS support to handle preflight OPTIONS requests

kowshik24:
Issue #284 , where the API failed to handle preflight OPTIONS requests, causing failures in direct client-to-API communications from browsers. The FastAPI application now includes CORSMiddleware, allowing proper handling of CORS preflight checks and enabling browser-based clients to directly interface with the Canopy API.

Changes:
- Added CORSMiddleware to the FastAPI application setup.
- Configured the middleware to allow all origins, methods, and headers, facilitating cross-origin requests.

This update ensures better accessibility and integration capabilities of the Canopy API for various client applications, especially those running in browser environments.

### DIFF
--- a/src/canopy_server/app.py
+++ b/src/canopy_server/app.py
@@ -23,6 +23,10 @@ from fastapi import (
     Body,
     APIRouter
 )
+
+from fastapi.middleware.cors import (
+    CORSMiddleware
+)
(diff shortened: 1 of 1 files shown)


miararoy (approved):
lgtm

PR #342, opened 2024-06-17
Added `source_tag='canopy'` to `knowledge_base._get_global_client()`

cwaddingham:
## Problem

It's not easy to quickly attribute index activity to Canopy, so we don't have a reliable way of comparing its use with other frameworks using Pinecone. 

## Solution

This adds a source_tag, 'canopy', which is inline with how our partners manage their attributions.

## Type of Change

- [ x] New feature (non-breaking change which adds functionality)

## Test Plan

1. Create a new Canopy index using the changed module.
2. Upsert one or more documents to the index.
3. Check the backend traces for the user_agent header "source_tag=canopy".


cwaddingham:
@miararoy from our conversation last week.

